PR #30, opened 2026-04-27
fix: hashing to avoid doc with duplicate name to collide

saccharin98:


saccharin98:
### Code review

No issues found. Checked for bugs (race conditions, data corruption, logic errors, edge cases) — no significant problems introduced by this PR.

Notes:
- The two potential concerns (parallel indexing race, raw/wiki name mismatch when file is already in raw/) were evaluated and determined to be false positives: the hash suffix prevents collisions in the normal code path, and lint.py's registry lookup correctly handles the naming indirection.

rejojer:
### Code review

saccharin98:
### Code review

KylinMountain:
### Code review

Found 1 issue:

1. Pre-existing registry entries (from versions before this PR) have no `doc_name`, `path`, `raw_path`, or `source_path` fields, so neither `get_by_path` nor `remove_by_doc_name` can match them. On re-ingest of a file whose content has changed, the upgrade path silently leaks orphans: the new hashed `doc_name` is generated, `remove_by_doc_name` matches nothing (old entry has no `doc_name` to compare), and the new entry is added alongside the stale one. The old `wiki/sources/<stem>.md` / `wiki/summaries/<stem>.md` files are also left in place next to the new `<stem>-<hash>.md` outputs. No migration code is included.

`remove_by_doc_name` only matches entries whose `metadata["doc_name"] == doc_name`:

https://github.com/VectifyAI/OpenKB/blob/870c947278865297d0072f082a13d93f824bf8d1/openkb/state.py#L55-L67

`get_by_path` only matches entries with the new path fields:

https://github.com/VectifyAI/OpenKB/blob/870c947278865297d0072f082a13d93f824bf8d1/openkb/state.py#L34-L44

The cleanup call in `add_single_file` therefore no-ops for any pre-PR entry:

https://github.com/VectifyAI/OpenKB/blob/870c947278865297d0072f082a13d93f824bf8d1/openkb/cli.py#L208-L223

A one-time migration (backfill `path`/`doc_name` for existing entries based on `name`) or a fallback match by `name` in `remove_by_doc_name` would close this gap.

KylinMountain:
Thanks for surfacing and fixing this, @saccharin98 — the duplicate-name collision was a real bug. We ended up addressing it in #96 with a slightly different approach: document identity is keyed by path, and a colliding same-stem source gets a deterministic `-{sha256(path)[:8]}` suffix (plus legacy-entry backfill and a full regression test suite). Since #96 covers this case, we're closing it in favor of #96. Thanks again for the original report and patch! 🙏

PR #42, opened 2026-05-07
Fix XLSX ingestion memory spikes with streaming parser

plasma16:
## Summary
- route `.xlsx` conversion through a streaming `openpyxl` reader (`read_only=True`, `data_only=True`)
- cap scan bounds (`max_rows=5000`, `max_cols=64`) to prevent pathological worksheet ranges from exploding memory
- stop scanning after sustained empty tails to avoid sparse-sheet runaway processing

## Why
Some workbooks report huge used ranges (e.g. `max_row=1048571`) despite having very little real data, which can cause generic converters to consume excessive RAM.

## Result
Significantly lower memory use during XLSX ingest while preserving useful sheet content for KB compilation.

### DIFF
--- a/openkb/converter.py
+++ b/openkb/converter.py
@@ -8,6 +8,7 @@ from pathlib import Path
 
 import pymupdf
 from markitdown import MarkItDown
+from openpyxl import load_workbook
 
 from openkb.config import load_config
 from openkb.images import copy_relative_images, extract_base64_images, convert_pdf_with_images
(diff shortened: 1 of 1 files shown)


KylinMountain:
### Code review

Found 1 issue:

1. `convert_xlsx_streaming` silently truncates spreadsheets at `max_rows=5000` and `max_cols=64` with no warning or signal back to the caller. When a sheet exceeds either bound, the excess is dropped, the function returns a partial markdown string, `convert_document` writes it to `wiki/sources/<doc_name>.md`, and the file hash is registered as fully processed — so the next `openkb add` will skip it as "already known." The user has no way to discover that a 50k-row workbook only ingested its first 5k rows. The early-exit on `empty_streak >= 200` has the same property. Every other branch in `convert_document` either converts the full document or raises.

The hard caps and the partial-success completion path:

https://github.com/VectifyAI/OpenKB/blob/c3a1f11e057376b99a1b76303d942eb65f778473/openkb/converter.py#L37-L88

Suggested fix: when the row/col cap is reached or `empty_streak` breaks the scan, append an explicit truncation marker to the returned markdown (e.g. `> Truncated: scan limits reached at row N / col M`), log a warning, and ideally surface a flag on `ConvertResult` so the caller can decide whether to register the hash as complete.

KylinMountain:
Thanks @plasma16, and sorry for the slow follow-up. The memory issue is real, but this no longer merges against `main` (the ingest/convert path was rewritten by the crash-safe mutation work in #142), and the 5,000-row cap truncates large sheets silently while still marking the source fully ingested — silent data loss in a knowledge base. We will handle XLSX memory separately with explicit truncation signaling. Closing — thanks for surfacing it.

PR #43, opened 2026-05-07
Incrementally resume long-PDF ingestion using cached PageIndex doc_id

plasma16:
## Summary
- add long-PDF ingest checkpoint state in `.openkb/long_pdf_jobs.json`
- cache `doc_id` and description after successful PageIndex indexing
- on re-run, reuse cached `doc_id` for long PDFs and retry compilation directly
- persist index/compile failure state for troubleshooting and incremental retry

## Why
When long PDF ingestion fails after indexing, re-running currently re-indexes the same document. This change makes retries incremental for long PDFs while leaving existing skip behavior unchanged for other file types.

## Scope
- only long-document (`long_pdf`) ingestion path
- no queue/cursor behavior for non-PDF files

### DIFF
--- a/openkb/cli.py
+++ b/openkb/cli.py
@@ -128,7 +128,7 @@ def _find_kb_dir(override: Path | None = None) -> Path | None:
     return None
 
 
-def add_single_file(file_path: Path, kb_dir: Path) -> None:
+def add_single_file(file_path: Path, kb_dir: Path) -> bool:
     """Convert, index, and compile a single document into the knowledge base.
 
     Steps:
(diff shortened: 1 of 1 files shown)


KylinMountain:
Thanks @plasma16. In the default local mode, PageIndex already content-hashes each file and reuses the cached `doc_id` without re-running the expensive parse/index pass, so local re-runs do not re-index from scratch; the only remaining gap is cloud-mode retries. The patch also no longer applies after #142 rewrote the function it targets. Closing for now — appreciate the dig into the resume path.

PR #44, opened 2026-05-07
Mitigate PageIndex file-descriptor buildup during long PDF indexing

plasma16:
## Purpose
Reduce risk of `[Errno 24] Too many open files` during long-PDF indexing with retries.

## Changes
- Add best-effort PageIndex client cleanup helper in `openkb/indexer.py`.
- Create a fresh `PageIndexClient` per retry attempt.
- Explicitly close backend storage after failed attempts and at function exit.
- Trigger `gc.collect()` between failed attempts to accelerate descriptor release in long runs.

## Why this helps
In local mode, repeated failed/retried indexing can leave resources alive longer than expected. This patch ensures resources are released promptly per attempt.

## Test evidence
- `python3 -m py_compile openkb/indexer.py` (pass)
- Full pytest suite not run in this environment because `pytest` is not installed in system Python.

## Notes
This is an OpenKB-side mitigation while upstream `pageindex` resource handling can be further hardened.

### DIFF
--- a/openkb/cli.py
+++ b/openkb/cli.py
@@ -191,7 +191,10 @@ def add_single_file(file_path: Path, kb_dir: Path) -> bool:
             click.echo(f"  Long document detected — indexing with PageIndex...")
             try:
                 from openkb.indexer import index_long_document
-                index_result = index_long_document(result.raw_path, kb_dir)
+                try:
+                    index_result = index_long_document(result.raw_path, kb_dir)
+                finally:
+                    _close_litellm_async_clients()
             except Exception as exc:
                 click.echo(f"  [ERROR] Indexing failed: {exc}")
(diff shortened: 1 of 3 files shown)


KylinMountain:
@plasma16 Thank you for your contribution. However, this PR introduces changes to other functionalities. Could you please scope this PR to only address the current feature? Thank you.

KylinMountain:
Thanks @plasma16 — good catch on the litellm async-client socket buildup. It was fixed independently on `main` in #95, which closes the cached async clients inside the compile loop `finally` (the same fix you were reaching for). The PageIndex/sqlite side is handled by the library itself (short-lived client per call + context-manager/`__del__` cleanup), and the PR no longer merges after the `main` rewrite. Closing.

PR #45, opened 2026-05-11
fix(query): make openkb query safe for non-TTY stdout (closes #34)

### LINKED ISSUE #34: run_query(stream=True) crashes on Windows when stdout is not a console (NoConsoleScreenBufferError from prompt_toolkit _fmt)
## Summary

  When `openkb.agent.query.run_query(..., stream=True)` runs with `sys.stdout` attached to anything other than a real Windows console (a pipe, a file,
   a captured subprocess stream, an MCP stdio transport), the streaming path calls `_fmt(...)` → `prompt_toolkit.shortcuts.print_formatted_text(...)`,
   which constructs a `prompt_toolkit.output.win32.Win32Output` and raises:

  ```
  prompt_toolkit.output.win32.NoConsoleScreenBufferError: No Windows console found. Are you running cmd.exe?
  ```

  This makes `openkb query` — which always passes `stream=True` — unusable from any non-interactive parent on Windows (subprocess wrappers, captured
  CI output, MCP servers, redirected output).

  ## Repro (Windows 11, openkb 0.1.3, Python 3.11)

  Either of these reproduces it; both detach stdout from a console:

  ```powershell
  openkb query "anything" *> output.log
  python -c "import subprocess; subprocess.run(['openkb','query','hi'], stdout=subprocess.PIPE, check=True)"
  ```

  The crash fires as soon as the agent emits its first tool-call line, at the `_fmt(...)` call site in `openkb/agent/query.py`.

  Setting `NO_COLOR=1`, `TERM=dumb`, `PYTHONIOENCODING=utf-8`, or `CREATE_NO_WINDOW` does **not** help — `prompt_toolkit` on Windows requires an
  actual console handle regardless of color/style settings.

  ## Root cause

  `run_query(stream=True)` in `src/openkb/agent/query.py` already computes the right gate at line 133:

  ```python
  use_color = sys.stdout.isatty() and not os.environ.get("NO_COLOR", "")
  ```

  …and correctly disables the Rich `Live` console when `use_color` is `False`. But the **prompt_toolkit `_fmt` path is not gated** — around line 211,
  `_fmt` is called unconditionally:

  ```python
  _fmt(style, ("class:tool", _format_tool_line(name, args) + "\n"))
  ```

  `_fmt` (`src/openkb/agent/chat.py:82`) is a thin wrapper around `print_formatted_text`, which on Windows constructs `Win32Output` and demands a real
   console handle even when no color attributes are being emitted. So even with `use_color = False`, this single line still crashes.

  ## Why it matters downstream

  Any process wrapping `openkb` programmatically on Windows hits this. We hit it building an MCP server that shelled out to `openkb query`; the
  failure is total (no answer ever returned), and `openkb query` exposes no flag to reach the clean `stream=False` branch. Our workaround is to bypass
   the CLI and call `openkb.agent.query.run_query(..., stream=False)` directly, which couples us to a non-public underscore-adjacent API.

  ## Suggested fix (smallest)

  Reuse the existing `use_color` check to also gate the prompt_toolkit/Rich output, e.g. around `query.py:211`:

  ```python
  if use_color:
      _fmt(style, ("class:tool", _format_tool_line(name, args) + "\n"))
  else:
      sys.stdout.write(_format_tool_line(name, args) + "\n")
      sys.stdout.flush()
  ```

  Or push the guard into `_fmt` itself in `chat.py:82` so every caller benefits — fall back to `sys.stdout.write(...)` when `not sys.stdout.isatty()`.
   Either way, no `prompt_toolkit` machinery should run when stdout isn't a TTY.

  ## Related (optional)

  `openkb query` always passes `stream=True`. Adding a `--no-stream` flag — or auto-disabling streaming when `not sys.stdout.isatty()` — would let
  non-interactive callers reach the clean non-streaming branch without library coupling.

  ## Environment

  - OS: Windows 11 Pro (10.0.26200)
  - Python: 3.11 (uv tool install)
  - openkb: 0.1.3
  - Reproduces under: `subprocess.run`/`Popen` with piped stdout, `> file` redirection, MCP stdio transport.

  Happy to send a PR if helpful — wanted to surface the diagnosis first.

Litash:
prompt_toolkit's print_formatted_text constructs a Win32Output on Windows that demands a real console handle, crashing run_query (stream=True) with NoConsoleScreenBufferError when stdout is a pipe, file, or captured subprocess stream (MCP stdio, redirected output, CI captures).

Two complementary changes:

1. Defensive: push the existing _use_color() gate into the _fmt helper in openkb/agent/chat.py so every prompt_toolkit call site -- chat REPL, query streaming, slash-command output -- falls back to plain sys.stdout.write when the output isn't a usable console.

2. UX: auto-disable stream=True in openkb query for non-TTY stdout via a small _stream_to_tty() helper in openkb/cli.py. Non-interactive callers now reach the clean stream=False branch in run_query and get the final answer string instead of an interleave of tool-call lines and answer tokens. The non-stream branch echoes the answer so output is preserved.

### DIFF
--- a/openkb/agent/chat.py
+++ b/openkb/agent/chat.py
@@ -80,6 +80,16 @@ def _build_style(use_color: bool) -> Style:
 
 
 def _fmt(style: Style, *fragments: tuple[str, str]) -> None:
+    # prompt_toolkit's print_formatted_text constructs a Win32Output on
+    # Windows that requires a real console handle, raising
+    # NoConsoleScreenBufferError when stdout is a pipe, file, or captured
+    # subprocess stream. Fall back to plain text when the output isn't a
+    # usable console.
+    if not _use_color(force_off=False):
+        for _, text in fragments:
+            sys.stdout.write(text)
+        sys.stdout.flush()
+        return
     print_formatted_text(FormattedText(list(fragments)), style=style, end="")
 
 
--- a/openkb/cli.py
+++ b/openkb/cli.py
@@ -10,6 +10,7 @@ warnings.filterwarnings("ignore")
 import asyncio
 import json
 import logging
+import sys
 import time
 from pathlib import Path
 
@@ -357,6 +358,17 @@ def add(ctx, path):
         add_single_file(target, kb_dir)
 
 
+def _stream_to_tty() -> bool:
+    """Return True when stdout is a real terminal.
+
+    Used to auto-disable streaming output when ``openkb query`` is piped,
+    redirected to a file, or run as a subprocess — streaming output emits
+    interleaved tool-call lines that are noisy for non-interactive callers,
+    and the non-streaming branch returns just the final answer string.
+    """
+    return sys.stdout.isatty()
+
+
 @cli.command()
 @click.argument("question")
 @click.option("--save", is_flag=True, default=False, help="Save the answer to wiki/explorations/.")
@@ -380,8 +392,11 @@ def query(ctx, question, save, raw):
     _setup_llm_key(kb_dir)
     model: str = config.get("model", DEFAULT_CONFIG["model"])
 
+    stream = _stream_to_tty()
     try:
-        answer = asyncio.run(run_query(question, kb_dir, model, stream=True, raw=raw))
+        answer = asyncio.run(run_query(question, kb_dir, model, stream=stream, raw=raw))
+        if not stream and answer:
+            click.echo(answer)
     except Exception as exc:
         click.echo(f"[ERROR] Query failed: {exc}")
         return
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -65,3 +65,51 @@ def test_init_already_exists(tmp_path):
         result = runner.invoke(cli, ["init"])
         assert result.exit_code == 0
         assert "already initialized" in result.output
+
+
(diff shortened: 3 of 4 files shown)


KylinMountain:
Reviewed. The fix matches the #34 diagnosis end-to-end and the two-layer defense is the right shape: `_fmt`'s plain-text fallback covers any path that still reaches `print_formatted_text`, while `_stream_to_tty()` keeps non-interactive callers out of the stream branch entirely. `--save` continues to work in both modes because `run_query` returns the answer string either way.

One small observation, not blocking:

The two `TestQueryStreamGate` tests patch `openkb.agent.query.run_query`, which works only because `cli.py` does `from openkb.agent.query import run_query` lazily inside the `query` command. If anyone ever hoists that import to module scope, the patch silently misses (cli binds `run_query` as a local name and the source-module patch doesn't touch it), and both tests would pass without exercising the new code. Worth either patching `openkb.cli.run_query` instead, or adding a brief comment so the next refactor doesn't quietly break the coverage.

Otherwise LGTM — clean fix, good test discipline (the `_boom` mock that raises on misuse is a nice touch).

KylinMountain (approved):
LGTM

PR #46, opened 2026-05-11
fix: stop silently dropping wiki updates when section/frontmatter drifts

### LINKED ISSUE #26: index.md file is not being updated when we call add folder
@BukeLy I was working on the main branch and tried to add a folder with some documents using `openkb add raw/<folder_name>`. The summaries, concepts and log.md are getting updated. However, I noticed that the `index.md` is not being updated and hence the `openkb chat` command has to go through all the indexed files under the `wiki` directory to answer the queries. 

However, this issue is not present when the package is being installed from pip. Can you please tag a specific commit from which the pip package was published? It will be of great help. Thanks!

KylinMountain:
## Summary

- Fixes #26: `openkb add <folder>` (and single-file add) appears to update `summaries/`, `concepts/`, and `log.md` but leaves `index.md` untouched when the user's `index.md` has drifted from the canonical layout.
- Same anti-pattern existed in three more places (`_backlink_summary`, `_backlink_concepts`, `_add_related_link`) — fixed in one pass.

## Root cause

Four helpers in `compiler.py` combined a **loose substring check** with a **strict newline-anchored `str.replace`**:

```python
if "## Documents" in text:                         # loose
    text = text.replace("## Documents\n", ...)     # strict
```

If the user's wiki had any small deviation — trailing space on a heading, CRLF line endings, hand-edited renaming, missing section, `sources:[a]` without a space after the colon — the if-branch matched but the replace did nothing, **silently dropping the entry**. No warning, no error, no log. The user's only signal was "why is my index.md never growing?"

That matches the symptom in #26 exactly: summaries/concepts/log keep updating (they're direct `write_text` / `append`), only `index.md` (which goes through this dance) silently fails.

## Changes

- New `_ensure_h2_section(lines, heading)` helper: auto-appends a missing H2 section and logs a `warning` so the drift is observable.
- New `_prepend_source_to_frontmatter(text, source_file)` helper: parses YAML frontmatter line-by-line and prepends to the inline `sources:` list, tolerating formatting variants.
- `_update_index`, `_backlink_summary`, `_backlink_concepts`, `_add_related_link` rewritten to use line-based section parsing via the existing `_get_section_bounds` / `_insert_section_entry` helpers instead of brittle string replace.

## Test plan

- All 220 existing tests still pass.
- 6 new regression tests covering:
  - `_update_index` with `## Documents` section missing
  - `_update_index` with `## Concepts` section missing
  - `_backlink_summary` / `_backlink_concepts` with trailing-whitespace heading
  - `_add_related_link` with `sources:[a]` (no space) and with no `sources:` line at all
- Total: 224 passed.

### DIFF
--- a/openkb/agent/compiler.py
+++ b/openkb/agent/compiler.py
@@ -312,21 +312,37 @@ def _read_concept_briefs(wiki_dir: Path) -> str:
     return "\n".join(lines) or "(none yet)"
 
 
+def _iter_h2_headings(lines: list[str]) -> list[tuple[int, str]]:
+    """Return ``[(line_index, normalized_heading), ...]`` for every ATX H2.
+
+    A line counts as H2 when it starts with ``"## "`` (two hashes + space).
+    ``normalized_heading`` is the line with trailing whitespace stripped, so
+    ``"## Documents "`` normalizes to ``"## Documents"`` — letting callers
+    use exact-string comparison without tripping on stray whitespace.
+
+    Used by ``_get_section_bounds`` so heading lookup and the next-section
+    boundary share one scan and one normalization rule.
+    """
+    return [
+        (i, line.rstrip())
+        for i, line in enumerate(lines)
+        if line.startswith("## ")
+    ]
+
+
 def _get_section_bounds(lines: list[str], heading: str) -> tuple[int, int] | None:
     """Return the [start, end) bounds for a Markdown H2 section.
 
-    Heading lookup tolerates trailing whitespace on the file's heading line
-    so a drifted ``## Documents `` still matches ``## Documents`` — otherwise
-    callers would treat the section as missing and append a duplicate H2.
+    Uses ``_iter_h2_headings`` so the same H2 detection that finds the
+    target heading also determines the section's end (the next H2). A
+    drifted ``"## Documents "`` matches ``"## Documents"`` because both
+    sides are normalized.
     """
-    for i, line in enumerate(lines):
-        if line.rstrip() == heading:
-            start = i + 1
-            end = len(lines)
-            for j in range(start, len(lines)):
-                if lines[j].startswith("## "):
-                    end = j
-                    break
+    headings = _iter_h2_headings(lines)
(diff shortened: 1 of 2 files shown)


KylinMountain:
### Code review

Found 1 issue:

1. The PR's stated goal is to recover from heading drift, but `_ensure_h2_section` relies on `_get_section_bounds`'s exact-equality match (`if line == heading`), so any drift like a trailing space on the heading is treated as "section missing" and a duplicate H2 is appended at the end of the file. Old entries stay stranded under the original (drifted) heading; new entries land under the freshly-appended one. Same family of silent corruption as the original bug, just shifted from "drop" to "duplicate."

The exact-equality check that defeats the recovery:

https://github.com/VectifyAI/OpenKB/blob/3c5d4c1a3e3535a41f4855d445cf3bcf6f1485e8/openkb/agent/compiler.py#L314-L327

The appender that runs whenever the exact match fails:

https://github.com/VectifyAI/OpenKB/blob/3c5d4c1a3e3535a41f4855d445cf3bcf6f1485e8/openkb/agent/compiler.py#L328-L349

The two new regression tests `test_section_with_trailing_whitespace_still_merges` pass despite this bug because they only assert `"[[concepts/transformer]]" in text` — the new link is present (in the appended clean section) but the file now has two `## Related Concepts` headings:

https://github.com/VectifyAI/OpenKB/blob/3c5d4c1a3e3535a41f4855d445cf3bcf6f1485e8/tests/test_compiler.py#L483-L498

Trace for input `"## Related Concepts \n- [[concepts/attention]]\n"` with `slugs=["attention","transformer"]`: `missing=["transformer"]` (attention already linked); `_ensure_h2_section` sees no exact `## Related Concepts` line → appends a new one → `_insert_section_entry` inserts `transformer` into the appended section. Output has two `## Related Concepts` H2s.

Suggested fix: change line 318 to `if line.rstrip() == heading:` (also tighten the new tests with `assert text.count("## Related Concepts") == 1`).

KylinMountain:
### Code review

PR #13, opened 2026-04-10
Simplify init prompts and capture API key to .env

rejojer:
## Summary
- Drop `language` and `pageindex_threshold` prompts from `openkb init`; both fall back to config defaults and can be edited later in `.openkb/config.yaml`.
- Add an interactive API key prompt (hidden input) that writes `LLM_API_KEY` to `./.env` (chmod 0600) when provided. Skips silently if left blank. If `.env` already exists, warn and don't overwrite.
- Polish the model prompt with provider/model examples for OpenAI / Anthropic / Gemini and a link to the LiteLLM providers page for others.

## Test plan
- [ ] `openkb init` in a fresh dir, accept default model, skip API key → `.env` should not exist.
- [ ] `openkb init`, accept default model, enter a fake key → `.env` exists with `LLM_API_KEY=...` and mode `0600`.
- [ ] `openkb init` when `.env` already exists, enter a key → warning printed, existing `.env` untouched.
- [ ] `openkb init` with a custom model (e.g. `anthropic/claude-sonnet-4-6`) → `.openkb/config.yaml` contains that model.

PR #48, opened 2026-05-14
feat(cli): add --language flag and prompt to openkb init

wooogy-dev:
## Summary

  - Add `--language/-l LANG` flag to `openkb init` so the wiki output language can be set non-interactively (e.g. `openkb   
  init --language ko`).
  - Add an interactive `Wiki language (enter for default en)` prompt that runs after the LLM API-key prompt when the flag is
   omitted; pressing enter accepts the configured default.                                                                  
  - Persist the chosen value to `.openkb/config.yaml` as `language:` instead of hard-coding `DEFAULT_CONFIG["language"]`.
                                                                                                                            
  ## Why                                                                             
                                                                                                                            
  Previously `openkb init` always wrote the default English language to `config.yaml`, so non-English users had to hand-edit
   the file right after initialization. Exposing it at init time — both via flag (for scripts / CI) and via prompt (for the
  interactive walkthrough) — removes that manual step.                                                                      
                                                                                     
  ## Test plan                                 

  - `tests/test_cli.py` — 7/7 pass:                                                                                         
    - `test_init_defaults_language_to_en` — no flag, default kept.
    - `test_init_language_flag_sets_config` / `test_init_language_short_flag` — `--language` / `-l` skip the prompt and     
  write the value.                                                                                                          
    - `test_init_language_prompt_accepts_input` — interactive path writes user input.
    - Pre-existing `test_init_creates_structure` / `test_init_schema_content` / `test_init_already_exists` updated to feed  
  prompt input.                                                                      
  - Full suite (`pytest`) — 224 passed.                                                                                     
  - Manual: `openkb init --language ko` and `openkb init -l Korean` in temp dirs → `.openkb/config.yaml` reflects the value;
   `openkb init --help` shows the flag.

### DIFF
--- a/openkb/cli.py
+++ b/openkb/cli.py
@@ -254,10 +254,56 @@ def use(path):
     click.echo(f"Default KB set to: {target}")
 
 
+_LANGUAGE_MAX_LEN = 50
+
+
+def _coerce_language(value: str | None) -> str | None:
+    """Strip a language string; treat blanks as unset; reject unsafe values.
+
+    The language string is interpolated into LLM system prompts (see
+    ``_SYSTEM_TEMPLATE`` in ``openkb/agent/compiler.py`` and the query agent's
+    instructions), so values with newlines or excessive length would let an
+    external caller smuggle instructions into the prompt. Capping at
+    ``_LANGUAGE_MAX_LEN`` and rejecting control characters is enough to close
+    that vector while still allowing common forms ("en", "ko", "Korean",
+    "Simplified Chinese").
+
+    Returns the cleaned string, or ``None`` if the input was missing or blank
+    after stripping. Raises ``click.BadParameter`` on unsafe input.
+    """
+    if value is None:
+        return None
+    value = value.strip()
+    if not value:
+        return None
+    if len(value) > _LANGUAGE_MAX_LEN or any(c in value for c in "\n\r\t"):
+        raise click.BadParameter(
+            f"language must be {_LANGUAGE_MAX_LEN} characters or fewer "
+            "with no control characters",
+            param_hint="'--language'",
+        )
+    return value
+
+
+def _language_option_callback(_ctx, _param, value):
+    return _coerce_language(value)
+
+
+def _stdin_is_tty() -> bool:
+    """Return True when stdin is a real terminal.
+
+    Used to skip optional ``openkb init`` prompts when input is piped or
+    redirected, so existing automation (e.g. ``printf '\\n\\n' | openkb init``)
+    keeps working as new prompts are added. Mirrors ``_stream_to_tty`` from #45.
+    """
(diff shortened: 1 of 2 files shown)


KylinMountain:
### Code review

KylinMountain:
A few smaller follow-ups from the same review pass that didn't make the main report — none blocking, but worth considering before merge:

**Empty `--language ""` bypasses the default fallback.** The guard `if language is None` does not catch an explicit empty string from `--language ""`, which then persists to `.openkb/config.yaml` as `language: ''` and renders as `"Write all content in  language."` (double space, broken) in [`compiler.py`](https://github.com/VectifyAI/OpenKB/blob/c318eac6e3e1ce2b3d4ce6e6dba37b9d56cb9c4d/openkb/agent/compiler.py#L44-L47) and [`query.py`](https://github.com/VectifyAI/OpenKB/blob/c318eac6e3e1ce2b3d4ce6e6dba37b9d56cb9c4d/openkb/agent/query.py#L45-L48). PR #13 already established the pattern for the `api_key` prompt — `.strip()` plus a truthiness check. One-liner fix: change the guard to `if not language` (or strip + re-prompt).

**Adding a 3rd prompt to `openkb init` is a silent break for existing automation.** Scripts piping `printf '\n\n' | openkb init` (two newlines for the model + api_key prompts) now `Abort!` on the new third prompt. The diff updates the tests' `input="\n\n"` to `"\n\n\n"` which confirms the break, but downstream CI scripts have no migration notice. Two options: (a) apply the `sys.stdin.isatty()` pattern from #45 so `init`'s prompts auto-default in non-TTY contexts, or (b) at minimum call this out in the PR description so users know to either pipe an extra newline or pass `--language en`. A parallel `--model` flag would also close the asymmetry where `init` is partially non-interactive.

**`--language` value flows verbatim into LLM system prompts.** A flag value like `--language "English. Ignore prior instructions..."` lands in the agent's system message unchanged. For a single-user KB this isn't a new attack surface (the user already owns their `config.yaml`), but if OpenKB is ever embedded into automation that passes external input to `--language`, this becomes a prompt-injection vector. A simple whitelist of common ISO 639-1 codes plus a length cap would close it without limiting practical use.

wooogy-dev:
@KylinMountain 

Addressed all three in 57d8ab6:

  1. `--language` callback strips and treats blank as unset, falling back
     to the default.
  2. Language prompt gated on `_stdin_is_tty()` (mirroring `_stream_to_tty`
     from #45); non-TTY uses the default, so `printf '\n\n' | openkb init`
     works again. `--model` left for a follow-up.
  3. `--language` rejects >50 chars or `\n`/`\r`/`\t` — blocks injection
     shapes while keeping `"Korean"` etc. usable.

KylinMountain (approved):
LGTM

PR #51, opened 2026-05-16
feat(cli): add `openkb remove` to safely delete a document (closes #41)

### LINKED ISSUE #41: Remove Document
Documents may get superceded with a different document or a higher version so we need a way to remove the document from the knowledge.

KylinMountain:
## Summary
- Adds `openkb remove <identifier>` for safely deleting a document from a knowledge base in one step, with a plan/confirm flow.
- Identifier resolves by exact filename → exact `doc_name` slug → fuzzy substring; ambiguous matches refuse to act.
- Concept pages whose only source was the removed doc are deleted by default; `--keep-empty-concepts` retains them with empty sources (useful when replacing the doc with a newer version). `--keep-raw` preserves the original file in `raw/`. `--dry-run` prints the plan only, `--yes` skips the confirm prompt.
- Auto-runs `lint --fix` after removal so stray wikilinks pointing at the removed summary or deleted concept pages get stripped automatically.

Closes #41.

## Test plan
- [x] `pytest` — full suite 287 passed (260 prior + 27 new in `tests/test_remove.py`).
- [x] End-to-end smoke test on a 2-doc / 3-concept KB:
  - single-source concept is deleted; multi-source concept is edited; unrelated concept untouched
  - `index.md` Documents + deleted-Concepts entries pruned, surviving entries kept
  - `--keep-raw` and `--keep-empty-concepts` flag paths verified
  - `--dry-run` prints plan without touching disk
  - dangling `[[summaries/...]]` link in a sibling page is stripped by the auto `lint --fix` pass
- [x] Ambiguous and unknown identifiers fail safe (no files modified).

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,6 +149,7 @@ A single source might touch 10-15 wiki pages. Knowledge accumulates: each docume
 |---|---|
 | `openkb init` | Initialize a new knowledge base (interactive) |
 | <code>openkb&nbsp;add&nbsp;&lt;file_or_dir&gt;</code> | Add documents and compile to wiki |
+| <code>openkb&nbsp;remove&nbsp;&lt;doc&gt;</code> | Remove a document and clean up its wiki pages, images, registry, and PageIndex state (use `--dry-run` to preview, `--keep-raw` / `--keep-empty-concepts` to retain artifacts) |
 | <code>openkb&nbsp;query&nbsp;"question"</code> | Ask a question over the knowledge base (use `--save` to save the answer to `wiki/explorations/`) |
 | `openkb chat` | Start an interactive multi-turn chat (use `--resume`, `--list`, `--delete` to manage sessions) |
 | `openkb watch` | Watch `raw/` and auto-compile new files |
--- a/openkb/agent/compiler.py
+++ b/openkb/agent/compiler.py
@@ -446,11 +446,13 @@ def _insert_section_entry(lines: list[str], heading: str, entry: str) -> bool:
 
 
 def _remove_section_entry(lines: list[str], heading: str, link: str) -> bool:
-    """Remove the first entry whose ``link`` appears in the named section.
+    """Remove the first entry whose line starts with ``- {link}`` in the named
+    section. Returns True if an entry was removed.
 
-    Returns True if an entry was removed. The match uses ``- {link}`` as
-    the prefix (the canonical bullet form that the other section helpers
-    produce); a fallback substring match catches custom bullet text.
+    Matching is intentionally strict (prefix-only, matching the canonical
+    bullet form written by ``_insert_section_entry`` and friends). An earlier
+    substring fallback could wrongly delete sibling bullets whose brief text
+    referenced the removed link.
     """
     bounds = _get_section_bounds(lines, heading)
     if bounds is None:
@@ -459,7 +461,7 @@ def _remove_section_entry(lines: list[str], heading: str, link: str) -> bool:
     start, end = bounds
     entry_prefix = f"- {link}"
     for i in range(start, end):
-        if lines[i].startswith(entry_prefix) or link in lines[i]:
+        if lines[i].startswith(entry_prefix):
             del lines[i]
             return True
     return False
@@ -750,8 +752,21 @@ def remove_doc_from_concept_pages(
             new_text = "\n".join(lines)
 
         # Drop standalone "See also: [[summaries/{doc_name}]]" lines.
+        # The dominant form (written by ``_add_related_link``) is a
+        # paragraph: preceded by a blank line and trailed by either a
+        # newline or end-of-string. The first regex matches that shape
+        # exactly, preserving one trailing newline so paragraph spacing
+        # in surrounding content survives.
         new_text = re.sub(
-            rf"^\s*See also:\s*\[\[{re.escape(bare_source)}\]\]\s*\n?",
+            rf"\n\n[ \t]*See also:[ \t]*\[\[{re.escape(bare_source)}\]\][ \t]*(\n|\Z)",
+            r"\1",
+            new_text,
+        )
+        # Fallback for hand-edited inline "See also:" lines that lack the
+        # paragraph-break separator above. Bounded to a single line via
+        # `[ \t]` and an optional trailing newline.
+        new_text = re.sub(
+            rf"^[ \t]*See also:[ \t]*\[\[{re.escape(bare_source)}\]\][ \t]*\n?",
             "",
             new_text,
             flags=re.MULTILINE,
--- a/openkb/agent/compiler.py
+++ b/openkb/agent/compiler.py
@@ -445,6 +445,26 @@ def _insert_section_entry(lines: list[str], heading: str, entry: str) -> bool:
     return True
 
 
+def _remove_section_entry(lines: list[str], heading: str, link: str) -> bool:
+    """Remove the first entry whose ``link`` appears in the named section.
+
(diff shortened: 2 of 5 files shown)


KylinMountain:
### Code review

Found 1 issue:

1. `HashRegistry.remove_by_doc_name` silently no-ops on real knowledge bases. `add_single_file` only persists `{"name", "type"}` to the registry — `doc_name` is never stored. The new `remove_by_doc_name` then iterates entries matching `meta.get("doc_name") == doc_name`, which is always None for KBs built via `openkb add`, so the hash entry never gets pruned. The plan still prints `REGISTRY remove hash entry (h_xxx…)`, but execution silently fails, and re-adding the same file later is wrongly skipped as "already known". Tests pass only because `_seed_two_doc_kb` manually injects `doc_name` into seed metadata.

   `add_single_file` registry write (missing `doc_name`): https://github.com/VectifyAI/OpenKB/blob/254f5a2474a3ee32d74e974ee9870c9a3b6d552e/openkb/cli.py#L209-L213

   `remove_by_doc_name` lookup that never matches: https://github.com/VectifyAI/OpenKB/blob/254f5a2474a3ee32d74e974ee9870c9a3b6d552e/openkb/state.py#L43-L52

   Fix: persist `doc_name` (and ideally `path`) in `add_single_file` (line 212), e.g. `registry.add(result.file_hash, {"name": file_path.name, "doc_name": doc_name, "type": doc_type, "path": str(result.raw_path.relative_to(kb_dir))})` — and add a regression test that seeds the registry via `openkb add` rather than direct injection.

KylinMountain:
### Code review

Functional completeness pass — does `openkb remove` clean up everything `openkb add` produces? Found 2 real gaps.

1. **Per-doc image directory is orphaned.** `openkb/images.py` writes every image into `wiki/sources/images/<doc_name>/` (the same `doc_name` stem used for the summary and source files), but the remove plan/executor never references that directory. For image-heavy PDFs / docx / pptx that's tens to hundreds of MB silently left behind per removed doc, with no `--keep-images` flag to even hint it was intentional.

   Plan only enumerates summary + sources + concepts + index + raw + registry: https://github.com/VectifyAI/OpenKB/blob/c504e26f3497de1da31bc8fb8677705ad7fbcc3c/openkb/cli.py#L519-L535

   Images directory written here, never matched on remove: https://github.com/VectifyAI/OpenKB/blob/c504e26f3497de1da31bc8fb8677705ad7fbcc3c/openkb/images.py#L65-L75

   Fix is small: compute `images_dir = wiki_dir / "sources" / "images" / doc_name`, add a `DELETE` action when it exists, and `shutil.rmtree(images_dir, ignore_errors=True)` in the executor block around line 612.

2. **PageIndex state is orphaned for long PDFs.** The long-doc ingest path calls `Collection.add(...)` and receives a `doc_id`, but the registry only persists `{name, doc_name, type}` — `doc_id` is dropped on the floor. `openkb remove` never imports `pageindex` and never calls `Collection.delete_document(doc_id)`, so for every removed long PDF this stays on disk: the SQLite row in `.openkb/pageindex.db`, the managed copy at `.openkb/files/<collection>/<doc_id>.pdf`, and any extracted images at `.openkb/files/<collection>/<doc_id>/images/`. For a 200MB PDF the entire 200MB managed copy leaks per add → remove cycle, and the document still shows up to anything that calls `col.list_documents()`.

   `doc_id` acquired here: https://github.com/VectifyAI/OpenKB/blob/c504e26f3497de1da31bc8fb8677705ad7fbcc3c/openkb/indexer.py#L54-L58

   Registry write drops `doc_id` — no handle survives to feed `delete_document`: https://github.com/VectifyAI/OpenKB/blob/c504e26f3497de1da31bc8fb8677705ad7fbcc3c/openkb/cli.py#L209-L216

   Fix is in two parts: (a) persist `doc_id` in the registry alongside `doc_name` for long-doc entries during ingest, (b) in the remove flow, if `meta.get("type") == "long_pdf"` and a `doc_id` is present, call `pageindex.Collection(storage_path=str(openkb_dir)).delete_document(doc_id)`. Existing KBs without `doc_id` will need a fallback (look up by basename via `col.list_documents()`) or a one-time migration.

Out of scope but worth noting (NOT flagged as bug): the body prose of multi-source concept pages is intentionally NOT regenerated on remove, and the `brief:` frontmatter stays as last synthesized. That's a reasonable design choice to avoid an LLM call per remove, but the remove command's help text doesn't say so — worth a one-line note in the docstring at `cli.py:470-482`.

KylinMountain:
### Code review

Logical-consistency pass — does the wiki end in a sound state for subsequent `add` after a `remove`? Found 1 issue.

1. **Failed PageIndex cleanup leaves a stale row that silently re-binds on re-add.** The executor removes the registry entry before attempting PageIndex cleanup. If the PageIndex call raises (LLM-key check inside `PageIndexClient.__init__` fails, network blip in cloud-LiteLLM lookup, transient SQLite lock, etc.), we WARN and continue — but the registry entry that held `doc_id` is already gone.

   On the next `openkb add <same file>`:
   - `convert_document` sees the hash is unknown (we cleared it) and proceeds.
   - The long-doc path calls `Collection.add(pdf_path)`. PageIndex's local backend dedupes by SHA-256 (`find_document_by_hash`) and returns the **stale `doc_id`** without re-parsing.
   - `col.get_document(stale_doc_id, include_text=True)` then feeds the OLD parsed structure to our compiler.
   - A fresh registry entry gets written pointing at the stale doc_id; the wiki is rebuilt on top of stale parse content.

   Net effect: a user who removes a document **because the parse was bad** and re-adds it gets the same bad parse back, with no signal anything went wrong. The inline rationale comment at the top of the PageIndex block says we deliberately put cleanup last "to leave only PageIndex bloat, not a half-removed wiki" — but the cost is worse than disk bloat: silent state divergence the user cannot recover from without manually inspecting `.openkb/pageindex.db`.

   Registry removed first: https://github.com/VectifyAI/OpenKB/blob/5e7613049bfbfc4bcfc041a39baab571e5411d81/openkb/cli.py#L694-L718

   Fix: reorder so `_cleanup_pageindex` runs before `registry.remove_by_doc_name`. On PageIndex failure the registry still holds `doc_id`, so `openkb remove` is retryable (PageIndex delete is idempotent — `find_document_by_hash` → row, `delete_document` unlinks file with `missing_ok=True` and `if doc_dir.exists()` guards). The reverse failure mode — PageIndex cleaned but registry remove fails — is also idempotent on retry.

Out of scope (didn't meet the 80-confidence threshold but worth tracking as follow-ups):
- `--keep-empty-concepts` retains stale concept body; re-adding the same doc enters the LLM "update" path and blends new content into prose still containing the old doc's contribution — likely to produce duplicate phrasing. (75)
- Multi-source concept pages may retain markdown image refs `alt` after `images_dir` is deleted; `fix_broken_links` only handles `[[wikilinks]]`. Real if the LLM echoes image refs into concept bodies — depends on actual concept content. (50)
- `PageIndexClient.__init__` validates LLM provider even for delete-only ops; users with rotated keys can't run the cleanup path and get a misleading "PageIndex cleanup failed" WARN. (75)

KylinMountain:
### Code review

The latest commit's executor reordering (PageIndex cleanup before the registry commit point) is sound: each pre-commit step is idempotent on retry (`unlink(missing_ok=True)`, `shutil.rmtree(ignore_errors=True)`, concept/index helpers that cheap-filter already-clean pages, `fix_broken_links` no-ops when nothing changes, and PageIndex's own `delete_document` handles missing file/dir gracefully), so the new two-test retry contract holds end-to-end.

PR #53, opened 2026-05-17
feat(cli): add `openkb feedback` to file a prefilled GitHub issue

KylinMountain:
## Summary

Adds `openkb feedback` — opens a prefilled GitHub issue in the user's browser. No backend, no maintainer-owned token, no `gh` CLI dependency, one flag.

```bash
openkb feedback                            # interactive (stdin)
openkb feedback "openkb add hangs"         # one-liner
openkb feedback --type bug "..."           # tags 'bug' (--type bug|feature|question|other)
echo "..." | openkb feedback               # piped / CI — type defaults to 'other'
```

The auto-attached **Diagnostics** block is intentionally tiny — `openkb` version, Python version, OS, and whether a KB is initialised in cwd. No paths, env vars, or keys.

## Why this shape

| Alternative | Problem |
|---|---|
| Auto-create via maintainer PAT | Secret in source / public client |
| Hosted feedback API | Backend + moderation + GDPR for an OSS CLI |
| User's `gh` CLI auth | Excludes anyone without `gh auth login`; also skips the natural "review before Submit" step that a browser flow gives |

Prefilled URL → browser is the standard pattern for "file a bug" CLI commands (rustup, cargo, etc.). GitHub renders the issue form with our title / body / label, the user reviews and clicks Submit with their own account.

## Behaviour details

- **TTY user**: prompts for `--type` if not given on the command line, then opens browser. URL is also printed for copy-fallback if the browser can't auto-open (headless box, no `$BROWSER`).
- **Non-TTY user** (CI / pipe / SSH): skips the type prompt, defaults to `other`. Browser open may fail silently — the printed URL is the fallback.
- **`webbrowser.open` returns False**: surfaced via stderr (`no browser available — copy the URL above`), exits 0. Tested via mock.

## Files

- `openkb/cli.py` — `feedback` command + 3 helpers (`_openkb_version`, `_collect_feedback_diagnostics`, `_build_feedback_url`)
- `tests/test_feedback.py` — 19 regression tests: URL shape, type → label mapping, title truncation/prefix rules, empty-dict diagnostics shape, the `webbrowser.open=False` headless path, the non-TTY `--type` skip path, version-helper consistency across cli/chat/`__init__`
- `README.md` — one row in the Commands table

## Self-review fixes applied during PR

After the initial commit I ran the code-review skill against my own PR and addressed 3 findings:

1. **`webbrowser.open` return value was discarded** — silent success on headless boxes. Now captured and reported.
2. **`_openkb_version` duplicated logic in 3 places with drifted fallback strings** — collapsed to one `from openkb import __version__; return __version__` that matches `openkb/agent/chat.py`.
3. **Type prompt wasn't TTY-gated** — same pattern PR #48 explicitly fixed; non-TTY would hang or abort. Now uses the existing `_stdin_is_tty()` helper to default to `other` in pipes / CI.

Then per discussion the original `--print-url` and `--no-diagnostics` flags were removed:
- `--print-url` was redundant — the URL is always printed for copy-fallback before `webbrowser.open` is attempted.
- `--no-diagnostics` was an over-correction — the diagnostics block is small enough and useful enough for maintainers that there's no real privacy reason to opt out.

Final surface: one flag (`--type`) plus the positional message argument.

## Test plan
- [x] 329 tests pass (310 prior + 19 new)
- [x] Smoke: `openkb feedback --type bug "msg"` opens browser with title/body/label correct (verified by mock + manual run)
- [x] Smoke: `echo "" | openkb feedback "X"` doesn't hang on the type prompt; URL is correct (manual run)
- [x] Version consistency: `cli._openkb_version()`, `chat._openkb_version()`, and `__init__.__version__` all return the same string (asserted in `test_openkb_version_helper_matches_package_version`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,7 +156,7 @@ A single source might touch 10-15 wiki pages. Knowledge accumulates: each docume
 | `openkb lint` | Run structural + knowledge health checks |
 | `openkb list` | List indexed documents and concepts |
 | `openkb status` | Show knowledge base stats |
-| <code>openkb&nbsp;feedback&nbsp;["msg"]</code> | File feedback by opening a prefilled GitHub issue (use `--type bug/feature/question`, `--print-url` for SSH / sandbox) |
+| <code>openkb&nbsp;feedback&nbsp;["msg"]</code> | File feedback by opening a prefilled GitHub issue (use `--type bug/feature/question` to tag the issue) |
 
 <!-- | `openkb lint --fix` | Auto-fix what it can | -->
 
--- a/README.md
+++ b/README.md
@@ -156,6 +156,7 @@ A single source might touch 10-15 wiki pages. Knowledge accumulates: each docume
 | `openkb lint` | Run structural + knowledge health checks |
 | `openkb list` | List indexed documents and concepts |
 | `openkb status` | Show knowledge base stats |
+| <code>openkb&nbsp;feedback&nbsp;["msg"]</code> | File feedback by opening a prefilled GitHub issue (use `--type bug/feature/question`, `--print-url` for SSH / sandbox) |
 
 <!-- | `openkb lint --fix` | Auto-fix what it can | -->
 
--- a/openkb/cli.py
+++ b/openkb/cli.py
@@ -1150,15 +1150,15 @@ _FEEDBACK_LABEL_MAP = {
 
 
 def _openkb_version() -> str:
-    """Return the installed openkb package version, or 'unknown' if it
-    can't be resolved (e.g. running from an editable checkout that isn't
-    on PYTHONPATH as a distribution).
+    """Return the installed openkb package version.
+
+    Delegates to ``openkb.__version__`` so the chat REPL, feedback issue
+    body, and any future caller all surface the same fallback string
+    (``0.0.0+unknown`` from ``openkb/__init__.py``). Mirrors
+    ``openkb.agent.chat._openkb_version``.
     """
-    try:
-        from importlib.metadata import version
-        return version("openkb")
-    except Exception:
-        return "unknown"
+    from openkb import __version__
+    return __version__
 
 
 def _collect_feedback_diagnostics(ctx) -> dict[str, str]:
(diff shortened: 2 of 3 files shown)


KylinMountain:
### Code review

Found 3 issues.

1. **`webbrowser.open` return value ignored — silent failure on headless boxes.** `webbrowser.open` returns `False` (no `Exception`) when no browser is available. On a CI runner, SSH session without GUI, or container with no `BROWSER` env, the command prints "Opening GitHub in your browser to file the issue." and exits 0 — but nothing happened. The fallback URL is printed before the call (good) so the user can copy it, but the success message lies. Fix: `if not webbrowser.open(url): click.echo("(no browser available — copy the URL above)", err=True)`.

   https://github.com/VectifyAI/OpenKB/blob/9ca027042906569196539043eedb11957955f95f/openkb/cli.py#L1273-L1283

2. **`_openkb_version()` is a third copy of the same logic, and the fallback string has already drifted.** The same `importlib.metadata.version("openkb")` pattern exists in `openkb/__init__.py` (fallback `"0.0.0+unknown"`) and is already wrapped by `openkb/agent/chat.py:112-114` as `_openkb_version()` that just delegates to `from openkb import __version__`. PR #53 adds a third re-implementation with fallback `"unknown"` — so an editable-install user submitting feedback gets a different version string than the same user in chat. Fix: replace the body with `from openkb import __version__; return __version__` to match `chat.py`.

   https://github.com/VectifyAI/OpenKB/blob/9ca027042906569196539043eedb11957955f95f/openkb/cli.py#L1150-L1163

3. **Interactive `--type` prompt isn't TTY-gated — repeats the silent-automation-break flagged on PR #48.** PR #48's review explicitly called this out: *"Adding a 3rd prompt to openkb init is a silent break for existing automation. Scripts piping printf '\n\n' | openkb init ... now Abort! on the new third prompt"* — and the accepted fix was the `_stdin_is_tty()` helper. PR #53 adds another `click.prompt(...)` at line 1256 with no TTY check, so `echo "msg" | openkb feedback` hangs (no more piped input for the type prompt) or aborts confusingly. Fix: gate with the existing `_stdin_is_tty()` helper and fall back to `--type other` (or require `--type` explicitly when stdin isn't a TTY).

   https://github.com/VectifyAI/OpenKB/blob/9ca027042906569196539043eedb11957955f95f/openkb/cli.py#L1254-L1264

Generated with Claude Code.

PR #57, opened 2026-05-18
feat(skill): distill an OpenKB wiki into a redistributable Anthropic Skill

KylinMountain:
## Summary

Adds a **Skill Factory** that distils an OpenKB wiki into a redistributable [Anthropic Skill](https://docs.claude.com/en/docs/build-with-claude/skills) — a portable directory that Claude Code, Codex CLI, Gemini CLI, and Cursor all install and load natively.

- New `openkb skill new <name> "<intent>"` CLI command writes `<kb>/output/skills/<name>/SKILL.md` (+ optional `references/<slug>.md` for depth-on-demand).
- New `/skill new` slash command in `openkb chat` — same generator, conversational entry point. The chat agent now has write access scoped to `<kb>/output/**` and `<kb>/wiki/explorations/**` so users can iterate via natural-language follow-ups.
- Auto-regenerated `<kb>/.claude-plugin/marketplace.json` lists all compiled skills, so pushing the KB to GitHub makes `npx skills@latest add <owner>/<repo>` work end-to-end.
- Companion commands: `openkb skill validate` (structural lint), `openkb skill eval` (trigger accuracy + body coverage), `openkb skill history` / `openkb skill rollback` (iteration safety).

## Architecture

Reuses the existing `openai-agents` SDK + LiteLLM stack from `openkb.agent.query`. All skill-related code lives under `openkb/skill/`:

- `creator.py` — builds the compile agent via `build_skill_create_agent`, system prompt loaded from `openkb/prompts/skill_create.md` and interpolated with intent + skill name + wiki schema. Tools are scoped: read wiki (`list_wiki_dir`, `read_wiki_file`, `get_page_content`, `get_image`), nested semantic search (`query_wiki`, narrow follow-ups only), write within the skill root.
- `generator.py` — the shared `Generator` primitive that wraps compile → validate → marketplace-publish so both CLI and chat entry points get the same quality gate.
- `validator.py` — pure-Python structural lint (frontmatter, name kebab-case, file sizes, references resolve, **foreign-wikilink gate** preventing `[[concepts/]]` / `[[summaries/]]` / `[[sources/]]` from leaking into the shipped artifact).
- `evaluator.py` — two-pass quality measurement: a trigger grader (description-only) and a coverage grader (body + references) run concurrently via `asyncio.gather` bounded by a semaphore. Surfaces ambiguous outputs and per-task failures separately so a single bad call doesn't discard the rest.
- `workspace.py` — `save_iteration` / `restore_iteration` with reversible-rollback safety: the current state is preserved as a new iteration before any overwrite.
- `marketplace.py` — regenerates the per-KB `.claude-plugin/marketplace.json`.
- `tools.py` — path-scoped IO primitives.

Shared helpers (`skill_dir`, `skills_root`, `skill_workspace_dir`, `extract_frontmatter`, `extract_body`, `extract_description`) live in `openkb/skill/__init__.py` as single sources of truth.

## Quality story

Three orthogonal gates catch different failure modes:

| Gate | Signal | When it runs |
|---|---|---|
| `validate_skill` | Structure: missing frontmatter, illegal description chars, foreign wikilinks in shipped output, oversized files | Auto, after every compile (inside `Generator.run`); also on demand via `openkb skill validate` |
| `run_eval` | **Trigger accuracy** (description-only grader fires on the right questions?) + **Body coverage** (body+references actually support what the description promises?) | On demand via `openkb skill eval` |
| `save_iteration` / `restore_iteration` | Iteration history: every overwrite preserves the prior version; rollback also saves the pre-rollback state first | Auto on `skill new -y`; on demand via `skill history` / `skill rollback` |

## Performance

- Skill compile uses `parallel_tool_calls=True` so the early read-fan-out phase batches independent reads in a single turn (~5-10 outer turns saved per compile, ~20-40s wall-clock).
- `run_eval` dispatches ~30 grader calls concurrently via `asyncio.gather` with `Semaphore(EVAL_CONCURRENCY=8)`. Empirical run on a 30-page PDF: skill compile **48s** end-to-end, eval **22s** for 20 prompts.
- Per-task failures captured with `return_exceptions=True` so one MaxTurnsExceeded doesn't discard the other 29 successful gradings.

## Test plan

- [x] `uv run pytest` — full suite green (500+ tests)
- [x] CLI: command name validation, KB detection, wiki-content gate, overwrite + iteration-save logic, marketplace.json regen, validate/eval/history/rollback flows
- [x] Chat `/skill new`: covered by dedicated tests; `write_kb_file` allow-list covered by 9 path-traversal tests
- [x] Validator: foreign-wikilink detection in body and references, frontmatter parse edge cases, file-size limits, kebab-case enforcement
- [x] Evaluator: trigger accuracy, body coverage, ambiguous grader output segregation, partial-failure isolation
- [x] Workspace: rollback preserves current state before overwrite (reversibility regression test)
- [x] **Manual acceptance**: ran `openkb skill new agent-design "Decide between workflows vs agents..."` against the Anthropic *Building Effective Agents* PDF. Result: trigger accuracy 19/20 (95%), body coverage 10/10 (100%), compile 48s, eval 22s. SKILL.md installed into Claude Code's `~/.claude/skills/` and verified to activate on composition-pattern questions.

## Known / deferred

- **Anthropic prompt caching** on the coverage grader: the openai-agents SDK takes `instructions` as a plain string with no hook for `cache_control` markers, so explicit Anthropic caching would require SDK-level work. OpenAI's automatic prefix caching already applies because the system prompt is byte-stable across the 10 coverage calls per eval.
- **Granularity**: one wiki source → one wide skill (e.g. `agent-design`) rather than fan-out to N narrow skills per source. Chosen deliberately — broad worldview skills are a valid Anthropic-Skills pattern when the description is tight and exclusionary. The validator's foreign-wikilink gate and the eval's body-coverage metric together protect against the failure modes that broad skills are prone to.


KylinMountain:
### Code review

Found 1 issue:

1. `MaxTurnsExceeded` is not caught — agent hitting `MAX_TURNS = 80` produces an unhandled traceback instead of a friendly error. `MaxTurnsExceeded` inherits from `AgentsException`, not `RuntimeError`, but both call sites (CLI and chat) only catch `RuntimeError`. Either widen the catch to `Exception` (or `AgentsException` + `RuntimeError`), or have `run_skill_compile` translate the SDK exception into a `RuntimeError` with a clear message.

https://github.com/VectifyAI/OpenKB/blob/746720098ccce057498840ae9f928e7f8350ef0e/openkb/agent/skill_compiler.py#L124-L131

Call sites that catch only `RuntimeError`:

https://github.com/VectifyAI/OpenKB/blob/746720098ccce057498840ae9f928e7f8350ef0e/openkb/cli.py#L1490-L1495

https://github.com/VectifyAI/OpenKB/blob/746720098ccce057498840ae9f928e7f8350ef0e/openkb/agent/chat.py#L540-L546

Generated with [Claude Code](https://claude.ai/code)

KylinMountain:
### Code review (design consistency)

Reviewed through the lens requested: 设计的一致性. Found 4 issues.

1. **`skill rollback` breaks the reversible-mutation invariant established by `skill new`.** `skill new` saves the current skill to `iteration-N/` before overwriting (workspace's whole reason for existing), but `restore_iteration` calls `shutil.rmtree(dest)` directly without first preserving the current state. A user who edits skill files in chat (`write_kb_file` allows this) and then runs `openkb skill rollback` loses those edits with no recovery path. Forward path is reversible; reverse path is not.

https://github.com/VectifyAI/OpenKB/blob/1679c8cedac512583003e0d18b0ef19b07cd84db/openkb/skill/workspace.py#L111-L120

2. **Quality gate diverges between CLI and chat entry points to the same generator.** `openkb skill new` auto-runs `validate_skill` after `gen.run()` and surfaces errors/warnings; `/skill new` in chat takes the same code path through `Generator.run()` but skips validation entirely and jumps to the success message. Same operation, different post-compile policy depending on where it was invoked.

https://github.com/VectifyAI/OpenKB/blob/1679c8cedac512583003e0d18b0ef19b07cd84db/openkb/agent/chat.py#L552-L560

https://github.com/VectifyAI/OpenKB/blob/1679c8cedac512583003e0d18b0ef19b07cd84db/openkb/cli.py#L1520-L1538

3. **`output/skills/<name>` is constructed independently in 6+ sites, with a 7th formula that's never consumed.** `creator.py`, `workspace.py` (`_skill_dir`), `marketplace.py` (`_list_skill_dirs`), `cli.py` (multiple subcommands), and `chat.py:533` each compute the path inline. `Generator.output_dir` introduces a *different* formula — `kb_dir / 'output' / f'{target_type}s' / name` — but both call sites discard `gen.run()`'s return value and rebuild the path themselves. The intended single source of truth is bypassed by every caller, so the new abstraction adds an unused fourth path-formula rather than collapsing the existing duplication.

https://github.com/VectifyAI/OpenKB/blob/1679c8cedac512583003e0d18b0ef19b07cd84db/openkb/skill/generator.py#L52-L67

https://github.com/VectifyAI/OpenKB/blob/1679c8cedac512583003e0d18b0ef19b07cd84db/openkb/skill/workspace.py#L29-L37

4. **SKILL.md frontmatter is parsed by four different implementations across the new package.** `validator.py` uses `splitlines()`+`.index('---')`; `evaluator.py` re-implements the same; `marketplace.py` uses a regex (`_FRONTMATTER_RE`/`_DESCRIPTION_RE`); `workspace.py` uses a regex plus a 2000-char prefix fallback. Four siblings parsing the same artifact, four edge-case behaviours. A description-parsing change in one will silently diverge from the other three.

https://github.com/VectifyAI/OpenKB/blob/1679c8cedac512583003e0d18b0ef19b07cd84db/openkb/skill/validator.py#L195-L210

https://github.com/VectifyAI/OpenKB/blob/1679c8cedac512583003e0d18b0ef19b07cd84db/openkb/skill/marketplace.py#L29-L40

https://github.com/VectifyAI/OpenKB/blob/1679c8cedac512583003e0d18b0ef19b07cd84db/openkb/skill/workspace.py#L125-L148

Lower-confidence consistency notes (not blocking): `skill validate`/`skill eval` drop the "Run \`openkb init\` first." hint that every other command emits; `/skill` slash-command description is the only lowercase entry in `_SLASH_COMMANDS`; `openkb/skill/tools.py` duplicates `openkb/agent/tools.py` with renamed-for-no-reason functions (`list_wiki_dir` vs `list_wiki_files`); stale comment at `cli.py:1468` still references `openkb/skill_workspace.py`.

KylinMountain:
### Code review (Skills domain quality)

Reviewed through the lens requested: 能否真的把一本书蒸馏成专家. The short answer is: today it produces a well-formed knowledge briefing, not an expert. Four structural reasons:

**1. `skill eval` measures description self-consistency, not skill quality.** Both `generate_eval_set` and `grade_one` receive only the skill's `description:` string — nothing reads the SKILL.md body, the references, or the wiki sources. A skill with a coherent description and a LOREM IPSUM body would score 100%. This is a closed loop, not a quality gate. Because the eval is marketed as how you verify the skill is good, this gap is load-bearing.

https://github.com/VectifyAI/OpenKB/blob/1679c8cedac512583003e0d18b0ef19b07cd84db/openkb/skill/evaluator.py#L87-L188

**2. The agent reads compressed wiki, never the raw source.** The compile prompt's working method directs the agent to read `wiki/concepts/*.md` and `wiki/summaries/*.md` — both of which are themselves LLM-synthesized compressions of the original document. There's no instruction to open `wiki/sources/`, and the anti-hallucination block even discourages it ("Do not copy large verbatim passages from `wiki/sources/`"). For a book input, the agent is distilling a summary of a summary; the specific examples, named techniques, edge cases, and counter-intuitive findings — the things that make a book worth distilling — get lost two compression hops above where the agent reads.

https://github.com/VectifyAI/OpenKB/blob/1679c8cedac512583003e0d18b0ef19b07cd84db/openkb/prompts/skill_create.md#L77-L96

**3. Granularity mismatch with the Anthropic Skills model.** Reference Anthropic skills are narrow operational tools — `pdf`, `skill-creator`, each with a tight trigger predicate and a procedural body. The PR's pipeline produces one skill per `openkb skill new` invocation, with no fan-out. The dogfooded `ai-native-startup-advisor` (in `output/skills/`) has a description spanning "idea, MVP, launch, and scale" — basically every question about running an AI startup. That description will over-trigger; that's the structural consequence of compiling a whole book into one skill rather than N narrow skills per book. The right unit of compilation is roughly one concept page per skill, which the pipeline doesn't support today.

https://github.com/VectifyAI/OpenKB/blob/1679c8cedac512583003e0d18b0ef19b07cd84db/openkb/skill/creator.py#L101-L154

**4. Source-grounded citation is a suggestion, not a gate.** The prompt instructs the agent to cite non-trivial claims with `[[wikilinks]]`, but the validator only checks that wikilinks *which appear* in the body resolve to real files — it does not require any wikilinks to be present. The team's own committed `skills/openkb/SKILL.md` contains zero body wikilinks. So a skill with hallucinated claims and no audit trail passes validation cleanly. For "expert distillation" framing, this is the difference between a citable specialist and a confident-sounding generalist.

https://github.com/VectifyAI/OpenKB/blob/1679c8cedac512583003e0d18b0ef19b07cd84db/openkb/skill/validator.py#L157-L195

**Verdict.** Drop a book in today and you reliably get: a structurally valid Markdown skill with a passable description, navigable wikilinks to whichever concept pages the agent picked, and `skill validate` green. You do not reliably get: the book's actual arguments encoded as decision rules, citation-grounded claims, or a trigger predicate narrow enough to avoid over-firing. The marketing claim ("Drop in a book. Out comes a digital expert.") is closer to the ceiling than the floor; the floor is "a well-organized research briefing."

The minimum changes that would close the largest part of the gap:
- Eval generator/grader read the SKILL.md body and references, not just the description, so a hollow body actually fails.
- Working method requires touching `wiki/sources/` (with the summary's `full_text` pointer) at least once per skill, not just summaries-of-summaries.
- A `skill improve` path that feeds the previous SKILL.md back as context so iteration accumulates rather than restarts.
- Require ≥1 body wikilink per N words; validator fails skills with zero source-grounded claims.

KylinMountain:
### Code review

Reviewed commit ce918cf. Found 1 issue:

1. `extract_description` is imported but never used in `openkb/skill/marketplace.py`. The function was added during the frontmatter-parser consolidation, but `_build_manifest` only emits fixed strings — no per-skill description interpolation happens here. Dead import.

https://github.com/VectifyAI/OpenKB/blob/ce918cfd70a68fac7fa2df0454bc4ca9f5e56263/openkb/skill/marketplace.py#L26-L30

Lower-confidence observations worth a glance (not blocking):

- `validator.py` runs `FOREIGN_WIKILINK_RE.findall(text)` on the full SKILL.md including frontmatter. If a `description:` value happens to contain a `[[concepts/...]]`-shaped literal, the gate fires with a "SKILL.md contains foreign wikilinks" message that points at the body, not the frontmatter. Behavior is arguably correct (a foreign wikilink in description is also a problem) but the error message is misleading. Either narrow the scan to the body or update the wording.

  https://github.com/VectifyAI/OpenKB/blob/ce918cfd70a68fac7fa2df0454bc4ca9f5e56263/openkb/skill/validator.py#L166-L189

- `evaluator.py` `grade_coverage` fail-closes to `unsupported` on ambiguous LLM output. The intent is honest (don't falsely claim coverage) but a garbled grader response now silently inflates `coverage_misses` with no trace, polluting `coverage_rate`. Consider a separate ambiguity counter or surfacing the raw output once for debugging.

  https://github.com/VectifyAI/OpenKB/blob/ce918cfd70a68fac7fa2df0454bc4ca9f5e56263/openkb/skill/evaluator.py#L318-L329

- `MaxTurnsExceeded` is imported inside `grade_coverage` only; the other two functions in the same module import it the same way. Not a circular-import workaround — could be lifted to module top for consistency and a small per-call savings.

KylinMountain:
### Code review (performance)

Reviewed b6607ec through 5 opus-model agents. Three issues ≥ 80 confidence:

1. **`run_eval` awaits 30 LLM calls sequentially** — the single largest wall-clock win in the PR. Trigger graders and coverage graders have zero state coupling between prompts, but the loop awaits each one before starting the next. For count=10 that's 20 trigger + 10 coverage = ~30 round-trips at ~1-2 s each = ~30-60 s. `await asyncio.gather(*[grade_one(...) for p in eval_set])` plus a separate `gather` for coverage collapses this to ~2 s — a **15-30× speedup**. Wrap in `asyncio.Semaphore(8)` to bound provider concurrency. ~10 LOC fix.

   https://github.com/VectifyAI/OpenKB/blob/b6607ecc830558801e6d7f0d37db2b28be2af749/openkb/skill/evaluator.py#L367-L387

2. **`grade_coverage` re-sends the same ~16 K body+references block 10 times with no prompt-caching marker.** For a 50 KB SKILL.md + 3×4 KB references, the coverage block is ~15.7 K input tokens, multiplied by 10 grader calls = ~157 K input tokens **per eval**. With provider prompt caching (Anthropic `cache_control` for 90% discount, or OpenAI's automatic ≥1024-token prefix caching), 9 of those 10 calls become cache hits. As coded, `skill eval` runs ~8× the input tokens of the pre-PR design. At gpt-4o pricing that's roughly **$0.48/eval vs $0.06**, enough that users will skip eval on cost grounds and lose the coverage signal the PR was designed to add.

   https://github.com/VectifyAI/OpenKB/blob/b6607ecc830558801e6d7f0d37db2b28be2af749/openkb/skill/evaluator.py#L297-L320

3. **Compile agent forces `parallel_tool_calls=False`.** The skill-create agent has a natural read-fan-out phase early in the compile — survey directories, read N summaries, follow `full_text` pointers to source pages. With `parallel_tool_calls=False` each `read_wiki_file` / `get_page_content` costs one full LLM round-trip. Allowing parallel tool calls lets the model batch independent reads, saving roughly 5-10 outer turns (~20-40 s per compile). Write phase serializes naturally because each `write_skill_file` depends on accumulated reads, so correctness risk is low.

   https://github.com/VectifyAI/OpenKB/blob/b6607ecc830558801e6d7f0d37db2b28be2af749/openkb/skill/creator.py#L141-L147

Lower-confidence observations worth a glance:

- `query_wiki` (creator.py) is a nested `Runner.run(max_turns=50)` inside the outer `max_turns=80`. Worst case is 80×50=4000 LLM calls before `MaxTurnsExceeded`. The "narrow follow-ups only" docstring is the only enforcement. Consider a per-compile call counter that returns an error after N invocations (e.g. 3).
- `_skill_content_block` is called twice per `run_eval` (once in `generate_eval_set`, once at the top of `run_eval`); each invocation re-reads SKILL.md and references. Trivial CPU/IO win but ~5 LOC to memoize.
- `restore_iteration` does `copytree → rmtree → copytree`. A `Path.rename(workspace_slot)` instead of the first `copytree` removes one full tree copy, ~10-50 ms on large skills.
- `REFERENCES_PREVIEW_BYTES = 4000` is per-file. A skill with 8 references already hits 32 KB before the body. A per-skill total budget would protect against ref-bloat.

KylinMountain:
### Code review

Reviewed commit 2b94014 (perf fixes). Found 1 issue:

1. **`query_wiki` tool docstring exposes the literal Python name `QUERY_WIKI_MAX_CALLS` instead of the value `3` to the model.** `@function_tool` captures the docstring at decoration time as the tool description visible to the LLM. The string isn't f-string-interpolated, so the model sees "Capped at QUERY_WIKI_MAX_CALLS invocations per compile" — meaningless to it. The agent can't self-regulate before hitting the cap; it just gets a surprise error string back. Either hardcode "3" in the docstring with a comment pointing at the constant, or rebuild the docstring after decoration so the model sees the actual number.

   https://github.com/VectifyAI/OpenKB/blob/2b9401413c17e885a2a42681703c351f055fbe10/openkb/skill/creator.py#L114-L130

Lower-confidence observations (non-blocking):

- `asyncio.gather` does not pass `return_exceptions=True`. A single grader raising `RuntimeError` (e.g. MaxTurnsExceeded on one prompt) aborts the whole eval and discards results from the other ~29 prompts that have already completed. Fail-fast is a defensible default for CLI tools, but on a 30-prompt eval that wastes a lot of completed work. Consider `return_exceptions=True` on the inner gathers with post-processing to keep partial results.

  https://github.com/VectifyAI/OpenKB/blob/2b9401413c17e885a2a42681703c351f055fbe10/openkb/skill/evaluator.py#L400-L403

- The evaluator module docstring's "Flow" section reads as a linear 1→5 sequence; it doesn't mention that steps 3-4 (the trigger and coverage graders) now run concurrently with a semaphore cap. Stale relative to the new code.

  https://github.com/VectifyAI/OpenKB/blob/2b9401413c17e885a2a42681703c351f055fbe10/openkb/skill/evaluator.py#L1-L30

- The `EVAL_CONCURRENCY = 8` comment claims "~4x the sequential baseline" — with default count=10 there are 30 grader calls, and both gather groups feed the same semaphore, so peak concurrency is 8 across the combined pool. The realistic speedup is closer to 30/8 ≈ 3.75× to 8× depending on provider latency variance, not a fixed ~4×.

  https://github.com/VectifyAI/OpenKB/blob/2b9401413c17e885a2a42681703c351f055fbe10/openkb/skill/evaluator.py#L49-L53

KylinMountain:
### Code review

Reviewed three commits since the last review (543edbe, 1b80b7f, b5c2512): the `query_wiki` cap revert plus two docstring/README cleanups. No issues found on the lines these commits touch.

The cap-removal worst-case theoretical bound (80 outer turns × 50 inner turns per `query_wiki` call) was acknowledged in the commit message as a deliberate policy choice with the fallback option of adding a logged warning if telemetry ever shows runaway compiles. The revert is surgical — `parallel_tool_calls=True`, `MAX_TURNS=80`, the tools list, and all other 2b94014 changes are untouched.

Notes from the prior review on 2b94014 remain non-blocking and not regressed by this set:
- `evaluator.py` module docstring "Flow" section still describes sequential steps where steps 3-4 actually run concurrently.
- `EVAL_CONCURRENCY = 8` comment still claims "~4x the sequential baseline"; the real speedup is closer to 30/8 across the shared pool.
- `asyncio.gather` still without `return_exceptions=True`; fail-fast remains the default.

PR #49, opened 2026-05-14
fix: eliminate ghost wikilinks in LLM-generated wiki content (closes #47)

KylinMountain:
## Summary

Closes #47. `openkb lint` was reporting 100+ broken `[[wikilinks]]` on every fresh ingest. The root cause is fully deterministic, not a legacy data artifact:

- `_compile_concepts` plan stage caps `create` at 2–3 concepts per doc by design (anti-proliferation).
- But `_CONCEPT_PAGE_USER` / `_SUMMARY_USER` told the LLM to "include wikilinks to related concepts" with no whitelist.
- LLM bodies ended up with 5–10 wikilinks each, most pointing to concepts the plan never created → ghost links.

Empirical evidence: the same 5 documents (PDF + xlsx + docx + 2 markdown) on the same `gpt-5.4-mini` configuration produce **115 broken wikilinks across 30 unique ghost targets** with the old code, **0 broken across 197 valid wikilinks** with this PR.

## Fix — three layers

**Layer A — prompt whitelist.** `_CONCEPT_PAGE_USER` and `_CONCEPT_UPDATE_USER` now receive `{known_targets}` (existing concepts on disk + `plan.create` + `plan.update` + `plan.related` + summary for this doc) and instruct the LLM to use `[[wikilinks]]` only for targets in the list, writing plain text otherwise.

**Layer B — in-memory deferred write + summary rewrite.** `compile_short_doc` no longer writes the v1 summary to disk immediately; it's held in memory and used as cache context for plan + concept generation. After all bodies are generated:
- Every concept body is run through `strip_ghost_wikilinks` (canonicalizes via NFKC + lowercase + `_↔-` normalization, drops unresolved links to plain text).
- The LLM regenerates the summary with the same cache-friendly message structure (BP1 + BP2 hit) but constrained to the full whitelist, and the v2 summary is also stripped before being written.

**Layer C — `lint --fix`.** `fix_broken_links` applies the same `strip_ghost_wikilinks` against existing wiki content so users with already-broken KBs can clean them up in place.

## Cache-control preserved

The summary-rewrite LLM call reuses the existing PR #38 cache breakpoints. Empirically:
- `summary-rewrite ... 6.8s (in=11887, out=869, cached=11008)` on the PDF doc → 92% of input tokens are cache hits.
- Marginal cost per doc: ~500–1000 fresh input tokens + ~1–2k output tokens (≈ 1/3 the cost of one concept body call).

## Compatibility with Karpathy's spec

[...]

PR #62, opened 2026-05-20
fix(cli): unbreak `openkb remove` — registry pruning + scoped lint pass (closes #58)

### LINKED ISSUE #58: openkb remove leaves orphan hash and reformats unrelated wiki
A single `openkb remove <doc>` run surfaces two independent bugs at once. Reporting them together because 
  they share a single repro, but they have different root causes and need separate fixes.                   
                                               
  Follow-up to #41 — both are regressions in the implementation shipped by PR #51.                          
                                                                                                          
  ## Repro                                                                                                  
   
  1. KB containing at least one document ingested before PR #51 (`hashes.json` entry has only `{name,       
  type}`, no `doc_name` key) and a handful of LLM-generated concept pages with pre-existing dangling      
  wikilinks.                                                                                                
  2. `openkb remove <that-doc>` (e.g. `openkb remove ollama`).                                            
  3. Observed: removal "succeeds" but `git status` / `hashes.json` show the symptoms below.
                                                                                                            
  ---
                                                                                                            
  ## Bug 1 — hash entry is not removed for docs ingested before PR #51                                            
                                               
  `cat .openkb/hashes.json` still contains the removed doc's entry after `openkb remove` reports success.   
  Re-running `openkb add <same-file>` is then incorrectly treated as a duplicate via the SHA dedup.       
                                                                                                            
  ### Root cause                                                                                          
                                               
  Commit c504e26 (within this same PR) fixed `add_single_file` so newly-ingested docs persist `doc_name`    
  into the registry. However, **entries that already existed in `hashes.json` before that commit were not 
  backfilled** — they still carry only `{name, type}`.                                                      
                                                                                                          
  `HashRegistry.remove_by_doc_name` (`openkb/state.py:44-51`) matches with `meta.get("doc_name") ==         
  doc_name`. For un-backfilled legacy entries the comparison evaluates to `None == "<slug>"` → always
  `False`. The method silently returns `False`; nothing in the call chain checks the return value.          
                                                                                                          
  Meanwhile `cli.py:670` (`doc_name = meta.get("doc_name") or Path(name).stem`) does fall back to the       
  filename stem to drive every other step, so summary/source/concept/index removal succeeds and the failure
  is invisible at the surface.                                                                              
                                                                                                          
  ### Suggested fix                            

  Either of the following — both are robust against un-backfilled legacy data:                              
  
  - Add a fallback in `remove_by_doc_name` that also matches when `Path(meta["name"]).stem == doc_name`, OR 
  - Introduce `remove_by_hash(file_hash)` and call it from `cli.py:842` since the CLI already has the     
  matched hash in hand. **Preferred** — eliminates the slug round-trip and works regardless of `doc_name`   
  presence.                                                                                               
                                                                                                            
  A one-shot migration that backfills `doc_name` on the next `openkb` invocation would also clean this up,  
  but the call-site fix above is sufficient and avoids touching user data on read paths.                                                                                         
                                                                                                            
  ---                                                                                                     
                                               
  ## Bug 2 — unrelated wiki pages get reformatted

  Removing a single doc produces a sprawling diff. In my repro, removing one `ollama.md` produced a         
  **39-file / 1254-line** diff; 27 of those were concept pages that didn't list ollama as a source.
                                                                                                            
  Example (from a concept page unrelated to ollama):                                                        
                                               
  ```diff                                                                                                   
  - **Knowledge access**: agents need curated context such as [[LLM Wiki]]                                
  + **Knowledge access**: agents need curated context such as LLM Wiki                                      
  ```
                                                                                                            
  ### Root cause                                                                                          
                                               
  `cli.py:815` calls `fix_broken_links(wiki_dir)` over the entire wiki on every remove.                     
  `openkb/lint.py:fix_broken_links` strips every dangling wikilink in the KB, not only the ones created by
  this removal. Pre-existing ghost links (LLM-generated, hand-edited, links to not-yet-added concepts, etc.)
   get swept up too.                                                                                      
                                               
  ### Impact

  - Removal commits are unreadable — actual deletion effects are buried under unrelated reformat noise.
  - Users lose `[[wikilinks]]` they may want to keep (e.g. links to a concept they plan to add later).
  - Violates least-surprise: the command name says "remove one doc," but the diff shows wiki-wide           
  refactoring.                                                                                              
                                                                                                            
  ### Suggested fix (preferred)                                                                             
                                                                                                          
  Limit ghost-link stripping to files actually touched by this removal: `concept_result["modified"]` ∪      
  `{index.md}`. Preserves the original PR #49 intent (clean up dangling links *the removal just created*) 
  without sweeping the rest of the KB.                                                                      
                                                                                                          
  ### Alternatives                             

  - Snapshot the global ghost set before/after the removal and strip only the *newly-introduced* ghosts.    
  - Make the global pass opt-in via a flag (e.g. `--lint`), default off.

wooogy-dev:
## Summary

Two regressions from #51 surface together on a single `openkb remove` run but have independent root causes — reported and fixed together because they share one repro, each landed as its own atomic commit.

Closes #58.

## Bug 1 — orphan hash entry on legacy documents

After `openkb remove` reports success, `hashes.json` still contains the removed doc's entry. Re-running `openkb add <same-file>` is then wrongly treated as a duplicate via SHA dedup.

**Root cause.** Commit c504e26 fixed `add_single_file` to persist `doc_name` into the registry, but rows ingested **before** that commit were never backfilled — they still carry only `{name, type}`. `HashRegistry.remove_by_doc_name` matched on `meta.get("doc_name") == doc_name`, which evaluated to `None == "<slug>"` and silently returned `False`. The CLI never checked the return value, so the failure was invisible at the surface.

**Fix** (commit 1). Prune by the already-resolved `file_hash` (returned by `_resolve_doc_identifier`) via a new `HashRegistry.remove_by_hash(file_hash) -> bool`. The hash key is always present, so pruning no longer depends on the metadata shape — and it drops the slug round-trip the CLI never needed.

## Bug 2 — `fix_broken_links` reformatting unrelated wiki pages

Removing a single doc produced a sprawling diff. In the repro, removing one `ollama.md` produced a **39-file / 1254-line** diff; 27 of those were concept pages that never listed `ollama` as a source.

**Root cause.** `fix_broken_links` was built for `openkb lint --fix` — an explicit, user-invoked wiki-wide cleanup — and strips *every* dangling `[[wikilink]]` in the KB. #51 reused it for `remove`'s post-pass, whose **declared intent (per the #51 PR description itself) was only to strip links pointing at the just-removed summary and deleted concept pages**. The post-remove cleanup intent was scoped; the helper it called was not. That mismatch swept the rest of the KB and silently demoted pre-existing ghost links the user meant to keep (e.g. a hand-written link to a concept they plan to add later).

To be precise: the wiki-wide behavior is *correct* for `openkb lint --fix`. This is a #51-side scoping fix, not a change to #49's behavior.

**Fix** (commit 2, the preferred path from the issue). Add a keyword-only `restrict_to` parameter to `fix_broken_links`. `remove` passes only the union of the concept pages it modified (`concept_result["modified"]`) and `index.md`, so the post-remove pass cleans only the dangling links *this remove just created*. The valid-target whitelist is still computed wiki-wide so legitimate `[[concepts/sibling]]` links in the scoped pages still resolve. `openkb lint --fix` is untouched — it omits the parameter and keeps the wiki-wide behavior.

## Test contract change

`test_cli_remove_lint_cleans_dangling_links` previously planted a stray link in `concepts/llm.md` (an unrelated page) and asserted it was stripped — that test enshrined the over-zealous behavior and couldn't tell scoped cleanup from global. Renamed to `..._in_modified_page` and re-targeted at `concepts/attention.md` (a page this removal actually modifies); the new `test_cli_remove_preserves_ghosts_in_unrelated_pages` locks in the opposite contract.

## Test plan

- [x] `tests/test_remove.py` — 46/46 pass, incl. `test_hash_registry_remove_by_hash`, `test_cli_remove_prunes_legacy_registry_entry_without_doc_name` (E2E Bug 1), `test_cli_remove_preserves_ghosts_in_unrelated_pages` (E2E Bug 2), `test_cli_remove_lint_cleans_dangling_links_in_modified_page`.
- [x] `tests/test_lint.py::TestFixBrokenLinksRestrictTo` — 5 new cases: default behavior, scoped, empty list = no-op, paths outside `wiki/` skipped, wiki-wide target set still consulted.
- [x] Full suite: **380 passed**, no regressions, no behavior change for `openkb lint --fix`.
- [x] **TDD verification.** With only the code patches reverted (tests applied), the 7 new regression assertions fail for the right reasons — `AttributeError` on `remove_by_hash`, `TypeError` on the `restrict_to` kwarg, the legacy `h_legacy` entry surviving, and `[[concepts/agent-loops]]` getting demoted to plain text — and the 2 sentinel tests pass on both sides of the patch.
- [x] Each commit verified in isolation (Bug 1 commit: 96/96 pass with Bug 2 changes stashed).

## Commits

1. `fix(cli): prune the hash registry by file_hash so openkb remove drops legacy entries` — Bug 1
2. `fix(cli): scope openkb remove's post-pass lint to the files it touched` — Bug 2

### DIFF
--- a/openkb/cli.py
+++ b/openkb/cli.py
@@ -839,7 +839,10 @@ def remove(ctx, identifier, keep_raw, keep_empty_concepts, dry_run, yes):
             return
 
     # ----- Commit point -----
-    registry.remove_by_doc_name(doc_name)
+    # Prune by hash, not by ``doc_name``: legacy registry entries
+    # (ingested before commit c504e26) carry only ``{name, type}`` and
+    # would silently no-op under ``remove_by_doc_name``. See issue #58.
+    registry.remove_by_hash(file_hash)
 
     if raw_path is not None:
         raw_path.unlink(missing_ok=True)
--- a/openkb/cli.py
+++ b/openkb/cli.py
@@ -812,7 +812,21 @@ def remove(ctx, identifier, keep_raw, keep_empty_concepts, dry_run, yes):
     # Strip dangling wikilinks now so a retry (after a PageIndex
     # failure below) finds a clean wiki — no point in re-running this
     # on every attempt.
-    files_changed, ghosts = fix_broken_links(wiki_dir)
+    #
+    # Scope: only the pages this remove actually touched (modified
+    # concept pages ∪ index.md). Previously this swept the whole wiki
+    # via ``fix_broken_links(wiki_dir)``, which silently stripped
+    # pre-existing dangling links in unrelated pages — see issue #58
+    # (Bug 2). Users who want a wiki-wide sweep can still run
+    # ``openkb lint --fix`` explicitly.
+    lint_scope: list[Path] = [
+        wiki_dir / "concepts" / f"{slug}.md"
+        for slug in concept_result["modified"]
+    ]
+    index_md = wiki_dir / "index.md"
+    if index_md.exists():
+        lint_scope.append(index_md)
+    files_changed, ghosts = fix_broken_links(wiki_dir, restrict_to=lint_scope)
     if files_changed:
         click.echo(f"  lint --fix cleaned {ghosts} dangling wikilink(s) in {files_changed} file(s)")
 
--- a/openkb/lint.py
+++ b/openkb/lint.py
@@ -173,7 +173,11 @@ def list_existing_wiki_targets(wiki_dir: Path) -> set[str]:
     return targets
 
 
-def fix_broken_links(wiki: Path) -> tuple[int, int]:
+def fix_broken_links(
+    wiki: Path,
+    *,
+    restrict_to: list[Path] | None = None,
+) -> tuple[int, int]:
     """Rewrite or strip broken [[wikilinks]] across the wiki in place.
 
     For each Markdown page under ``wiki`` (excluding ``reports/`` and
@@ -184,6 +188,16 @@ def fix_broken_links(wiki: Path) -> tuple[int, int]:
 
     Args:
         wiki: Path to the wiki root directory.
+        restrict_to: When provided, only rewrite these files (must live
+            under ``wiki``). Paths outside the wiki and non-existent
+            paths are silently skipped. An empty list is a no-op — the
+            valid-target whitelist is still computed from the entire
+            wiki, so links like ``[[concepts/sibling]]`` resolve
+            correctly even when ``sibling.md`` is not in the scope.
+            Used by ``openkb remove`` (issue #58 / Bug 2) to clean only
+            the pages it actually touched instead of sweeping the
+            whole wiki and stripping pre-existing dangling links the
+            user may want to keep.
 
     Returns:
         Tuple of ``(files_changed, ghosts_stripped)``.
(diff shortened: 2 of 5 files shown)


KylinMountain:
### Code review

KylinMountain:
Thanks @SeungwookHan  — great repro, clean root-cause analysis, and splitting the two regressions into separate commits made this a pleasure to review. 

The TDD verification note (running the new tests against reverted patches to confirm they fail for the right reasons) is exactly the level of rigor I want on regression fixes. Merging.

KylinMountain (approved):
LGTM

wooogy-dev:
@KylinMountain 
Thanks! Glad the split made it a pleasure to review — happy to revisit the scope boundary if needed.

PR #65, opened 2026-05-23
fix(deps): pin markitdown >=0.1.5 with narrow extras (closes #64)

### LINKED ISSUE #64: Error converting pptx file with charts
`openkb add PATH` fails for a pptx file with the following error. The process is not continued with the next file but aborted completely.

Traceback:
```
Traceback (most recent call last):
  File "PATH/.venv/lib/python3.14/site-packages/markitdown/_markitdown.py", line 1715, in _convert
    res = converter.convert(local_path, **_kwargs)
  File "PATH/.venv/lib/python3.14/site-packages/markitdown/_markitdown.py", line 903, in convert
    md_content += self._convert_chart_to_markdown(shape.chart)
                  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^
  File "PATH/.venv/lib/python3.14/site-packages/markitdown/_markitdown.py", line 952, in _convert_chart_to_markdown
    row.append(series.values[idx])
               ^^^^^^^^^^^^^
  File "PATH/.venv/lib/python3.14/site-packages/pptx/chart/series.py", line 83, in values
    return tuple(iter_values())
  File "PATH/.venv/lib/python3.14/site-packages/pptx/chart/series.py", line 81, in iter_values
    yield val.pt_v(idx)
          ~~~~~~~~^^^^^
  File "PATH/.venv/lib/python3.14/site-packages/pptx/oxml/chart/series.py", line 103, in pt_v
    return results[0].value if results else None
           ^^^^^^^^^^^^^^^^
  File "PATH/.venv/lib/python3.14/site-packages/pptx/oxml/chart/series.py", line 254, in value
    return float(self.v.text)
ValueError: could not convert string to float: '#N/A'
```

KylinMountain:
## Summary

- Replace `markitdown[all]` with `markitdown[docx,pptx,xlsx,xls]>=0.1.5` — drops ~67 MB of unused deps (azure-*, pdfminer, pdfplumber, speechrecognition, youtube-transcript-api, pydub, xlrd-then-readded, olefile) and forces users off pre-0.1.0 markitdown.
- Add `.xls` to `SUPPORTED_EXTENSIONS` / `_SHORT_DOC_TYPES` so the `[xls]` extra is actually reachable.

## Why this closes #64

The reporter's traceback (`_markitdown.py:1715`) is from the pre-0.1.0 monolithic markitdown, where a `ValueError` from python-pptx's `CT_StrVal_NumVal_Composite.value` (bare `float(self.v.text)` on `#N/A` / `#DIV/0!` / blank cells) propagates straight out and aborts the whole `openkb add` run.

From markitdown 0.1.2 onward, `_convert_chart_to_markdown` is wrapped in `except Exception` and the offending chart degrades to `[unsupported chart]`. Pinning `>=0.1.5` guarantees users land on a version with that handler.

## Caveat

This is an "unblock the pipeline" fix, not a chart-data-preservation fix — the chart's other (numeric) cells are still lost on bad cells. The root cause sits in python-pptx and there is no upstream fix or open PR there; we considered a monkey-patch but dropped it as too invasive. A higher-fidelity pptx path (e.g. LibreOffice → PDF → existing PDF pipeline) behind a config flag is a reasonable follow-up if users hit this in practice.

## Test plan

- [x] `uv run --extra dev pytest` → 520 passed
- [x] `uv sync` after the dep change cleanly removes 17 packages, then re-adds `xlrd` for `[xls]`
- [x] Manual: `from openkb.converter import MarkItDown` still imports

### DIFF
--- a/openkb/cli.py
+++ b/openkb/cli.py
@@ -92,7 +92,7 @@ def _setup_llm_key(kb_dir: Path | None = None) -> None:
 
 # Supported document extensions for the `add` command
 SUPPORTED_EXTENSIONS = {
-    ".pdf", ".md", ".markdown", ".docx", ".pptx", ".xlsx",
+    ".pdf", ".md", ".markdown", ".docx", ".pptx", ".xlsx", ".xls",
     ".html", ".htm", ".txt", ".csv",
 }
 
@@ -101,7 +101,7 @@ _TYPE_DISPLAY_MAP = {
     "long_pdf": "pageindex",
 }
 
-_SHORT_DOC_TYPES = {"pdf", "docx", "md", "markdown", "html", "htm", "txt", "csv", "pptx", "xlsx"}
+_SHORT_DOC_TYPES = {"pdf", "docx", "md", "markdown", "html", "htm", "txt", "csv", "pptx", "xlsx", "xls"}
 
 
 def _display_type(raw_type: str) -> str:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,7 +27,7 @@ classifiers = [
 keywords = ["ai", "rag", "retrieval", "knowledge-base", "llm", "pageindex", "agents", "document"]
 dependencies = [
     "pageindex==0.3.0.dev1",
-    "markitdown[all]",
+    "markitdown[docx,pptx,xlsx,xls]>=0.1.5",
     "trafilatura>=2.0",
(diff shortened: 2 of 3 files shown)


KylinMountain:
### Code review

Issue #71, opened 2026-05-25
compiler: concept generation fails with  and JSON parse error on same document


moyu12-ae:
## Description

When adding a Chinese EPUB-to-Markdown document (《给青年编剧的信3.0》, ~80k chars), the compiler fails to generate concept pages in two different ways across two attempts, while other documents process fine.

## Environment
- openkb v0.1.dev1+g51b6d4f88
- Python 3.12
- macOS 24.2.0 (darwin)
- Model: deepseek-v4-pro (via LiteLLM)

## Steps to Reproduce

1. Convert an EPUB to Markdown via pandoc: `pandoc book.epub -t markdown -o raw/book.md`
2. Add to knowledge base: `openkb add raw/book.md`
3. Observe compiler warnings

## First attempt — `'list' object has no attribute 'get'`

The LLM successfully returned a concepts plan (3 concepts + 1 update), but all 3 concept generations failed:

```
Compiling short doc...
  summary.............................. 23.6s (in=80249, out=1149)
  concepts-plan................ 16.3s (in=81675, out=733, cached=80187)
  Generating 4 concept(s) (concurrency=5)...
  concept: 剧作中心制... 18.3s (in=81423, out=1062, cached=80635)
  concept: 喜剧创作... 18.7s (in=81417, out=924, cached=80635)
  concept: 钩子与阻力... 24.2s (in=81420, out=1287, cached=80635)
  update: dramatic-conflict... 35.0s (in=82579, out=2483, cached=80635)
openkb.agent.compiler WARNING: Concept generation failed: 'list' object has no attribute 'get'
openkb.agent.compiler WARNING: Concept generation failed: 'list' object has no attribute 'get'
openkb.agent.compiler WARNING: Concept generation failed: 'list' object has no attribute 'get'
```

Root cause appears to be in `compiler.py` line 946-947: when `_parse_json` returns a list, the fallback treats it as a flat dict list via `plan = {"create": parsed, ...}`, but if any create item is itself a list instead of a dict, line 1002 (`concept.get("title", name)`) raises `'list' object has no attribute 'get'`:

```python
# compiler.py:946-947
if isinstance(parsed, list):
    plan = {"create": parsed, "update": [], "related": []}
```

## Second attempt — `Failed to parse concepts plan: JSON parse error`

After removing and re-adding, the plan JSON couldn't parse at all:

```
Compiling short doc...
  summary.............................. 30.4s (in=80249, out=1440)
[...]

PR #89, opened 2026-06-09
fix(compiler): unwrap JSON-array LLM responses so concept/entity pages aren't silently dropped

cnndabbler:
## Problem
The four page generators in `compiler.py` (`_gen_create`, `_gen_update`, `_gen_entity_create`, `_gen_entity_update`) do:

```python
parsed = _parse_json(raw)
brief = parsed.get("brief", "")
```

When the model returns a **JSON array** instead of an object, `_parse_json` returns a `list`, and `parsed.get(...)` raises `AttributeError: 'list' object has no attribute 'get'`. That error is **not** caught by the surrounding `except (json.JSONDecodeError, ValueError)`, so the whole coroutine dies and the page is silently dropped:

```
openkb.agent.compiler WARNING: Entity generation failed: 'list' object has no attribute 'get'
[WARN] N entity(ies) planned but only M written ... (AttributeError).
```

This is the `'list' object has no attribute 'get'` failure reported in #71. It still reproduces on current `main` (observed across a 165-doc ingest with several models — deepseek, qwen — affecting ~20% of docs, each losing one or more pages).

## Fix
Add `_as_obj()` next to `_parse_json()`: if the parsed value is a list, unwrap the first dict element (the common "model wrapped the object in a one-element array" case); otherwise raise `ValueError` so the callers' existing raw-body fallback applies. Wrap all four call sites as `parsed = _as_obj(_parse_json(raw))`.

No behavior change for the normal object case.

## Tests
Adds `TestAsObj` covering dict passthrough, single-element unwrap, first-dict-in-list, and the list-without-dict / empty-list `ValueError` paths. Full suite green locally.

Relates to #71.

### DIFF
--- a/openkb/agent/compiler.py
+++ b/openkb/agent/compiler.py
@@ -393,6 +393,23 @@ def _parse_json(text: str) -> list | dict:
     return result
 
 
+def _as_obj(parsed: list | dict) -> dict:
+    """Coerce a parsed JSON result to a dict.
+
+    The page/entity generators expect a JSON object, but the LLM sometimes
+    wraps that single object in a one-element array. Unwrap the first dict
+    element so ``.get(...)`` doesn't blow up with AttributeError on a list.
+    Raises ValueError when no dict is recoverable, so callers' existing
+    ``except (json.JSONDecodeError, ValueError)`` falls back to the raw body.
+    """
+    if isinstance(parsed, dict):
+        return parsed
+    for item in parsed:
+        if isinstance(item, dict):
+            return item
+    raise ValueError("Expected JSON object, got list with no dict element")
+
+
 def _filter_concept_items(items: list, label: str) -> list[dict]:
     """Keep only dicts that carry a non-empty ``name``; warn about anything else."""
     if not isinstance(items, list):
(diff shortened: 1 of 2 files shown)


KylinMountain:
Nice, clean fix and good test coverage — `_as_obj` is correct and the four call sites correctly fall back via the existing `except (json.JSONDecodeError, ValueError)`.

One thing before merge: **the fix is incomplete** — the same `.get`-on-a-`list` crash still lives in the summary step of `compile_short_doc`, which this PR does not wrap:

https://github.com/VectifyAI/OpenKB/blob/1ad818936530c3cc101bde353ea984ab635b8fa0/openkb/agent/compiler.py#L1992-L1999

```python
summary_parsed = _parse_json(summary_raw)
doc_brief = summary_parsed.get("brief", "")          # AttributeError when the model returns a JSON array
summary = summary_parsed.get("content", summary_raw)
except (json.JSONDecodeError, ValueError):           # does not catch AttributeError
```

It uses the same `_parse_json`, the same models, and the same `_JSON_RESPONSE_FORMAT`, so the array-wrapping case from #71 is just as likely here. And the blast radius is larger: the uncaught `AttributeError` propagates out of `compile_short_doc`, killing the whole short-doc compile rather than dropping a single page. Suggest applying the same `_as_obj(_parse_json(summary_raw))` here.

(For reference, the plan-parse site is already safe — it guards with an `isinstance(parsed, (list, dict))` check before calling `.get`.)

Minor / optional:
- `_as_obj` silently discards all but the first dict when the list has multiple objects. Fine for the one-element wrap case, just noting it.
- Since five sites now want "parse to a dict", a small `_parse_json_obj(raw)` wrapper would dedupe the call sites and naturally cover the summary path so it cant be missed again.

PR #91, opened 2026-06-09
fix(cli): close litellm async clients after each compile/index (CLOSE-WAIT/FD leak)

cnndabbler:
## Problem
`add_single_file` compiles each document in a fresh `asyncio.run()` event loop. LiteLLM caches its async (aiohttp) clients per event loop, so when each loop ends the previous doc's clients are abandoned without being closed. Their HTTP connections sit in `CLOSE-WAIT` and **accumulate sockets/file descriptors across a long ingest**.

Observed on a 165-document ingest against a remote API: the process held **200+ sockets in CLOSE-WAIT**, climbing per doc. (On a box with a low `ulimit -n` this would eventually exhaust FDs and start failing compilations.)

## Fix
Add a best-effort `_close_litellm_async_clients()` (calls litellm's own `close_litellm_async_clients()`, never raises) and invoke it in `try/finally` around the three async entry points in `add_single_file`:
- `index_long_document(...)`
- `asyncio.run(compile_long_doc(...))`
- `asyncio.run(compile_short_doc(...))`

So cached clients are closed after each doc whether it succeeds or fails.

## Verification
Added a doc end-to-end after the change: `CLOSE-WAIT` returns to ~0 after each doc instead of accumulating. Updated `test_add_short_doc_runs_compiler` (the compile path now drives `asyncio.run` for both the compile and the cleanup, so it asserts the `compile_short_doc` coroutine was run rather than that `asyncio.run` was called exactly once).

## Relation to #44
#44 carried this same intent but is now ~23 commits behind `main` and conflicts with the current `indexer.py`/`cli.py`. This is a minimal reimplementation on current `main`, so it supersedes #44.

### DIFF
--- a/openkb/cli.py
+++ b/openkb/cli.py
@@ -259,6 +259,20 @@ def _clear_existing_skill_dir(kb_dir: Path, name: str) -> None:
         shutil.rmtree(target)
 
 
+def _close_litellm_async_clients() -> None:
+    """Best-effort cleanup of cached LiteLLM async clients.
+
+    LiteLLM caches aiohttp clients per event loop; with ``asyncio.run`` creating
+    a fresh loop per doc, the old clients' connections linger in CLOSE-WAIT and
+    accumulate sockets/FDs over a long ingest. Closing them after each
+    compile/index frees those connections. Cleanup must never break ingest.
+    """
+    try:
+        asyncio.run(litellm.close_litellm_async_clients())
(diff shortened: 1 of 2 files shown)


KylinMountain (approved):
LGTM. Thank you so much for your contribution — this is a great fix. I'll go ahead and merge your PR.

PR #104, opened 2026-06-18
feat: concurrent add pipeline with crash-safe mutation recovery

gwokhou:
## Summary

Parallelize `openkb add <dir>` and make every KB mutation crash-recoverable through a journaled snapshot/rollback layer.

## Why

Directory `add` was strictly serial — ingest time scaled linearly with file count and was dominated by conversion. Worse, a crash between convert and the registry write orphaned raw/source artifacts in the live KB with no way to recover or clean them up.

## Changes

- **Concurrent prepare, serialized commit.** File preparation (hash, duplicate prefilter, staging, conversion) runs on a `ThreadPoolExecutor` (`file_processing_jobs`, default 2); live-KB mutation — publish, PageIndex indexing, LLM compile, registry write, log append — stays under the mutation lock. Commits run in scan order for stable `log.md` / CLI output.
- **Crash-safe mutation journal** (`openkb/mutation.py`). Each commit snapshots the KB paths it will touch, journals the intent, and rolls back on failure. `recover_pending_journals` rolls back any `active` journal left by an interrupted run and discards `committed`/`rolled_back` ones. SQLite sidecars (`pageindex.db-wal/-shm/-journal`) are snapshotted too, so long-doc failures don't leave sidecars newer than the db.
- **Recovery wired into the lock.** Draining runs on every exclusive-lock acquisition, so `remove`/`recompile`/`lint`/`chat` also restore the KB before mutating — not just `add`. Prevents a crashed `add`'s journal from later clobbering an intervening `remove`.
- **Staged conversion.** `convert_document` writes into an isolated staging dir published atomically at commit, so a crash can no longer orphan live-KB artifacts.
- **Perf & diagnostics.** Prefilter known hashes before preparing; DEBUG-level per-stage timing logs.

## Tests

`test_mutation.py` + `test_add_command.py` cover snapshot rollback, commit protection, partial-failure cleanup, lock-driven recovery, and the jobs>1 pipeline end-to-end (real prepare + commit, only LLM compile mocked). Full suite: 820 passed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -348,6 +348,8 @@ Model names use `provider/model` LiteLLM [format](https://docs.litellm.ai/docs/p
 <summary><i>Advanced options (entity_types, extra_headers, OAuth):</i></summary>
 <br>
 
+`file_processing_jobs` (default `2`): number of files prepared concurrently during `openkb add <dir>`. Only the preparation stage is parallelized (hashing, duplicate prefiltering, raw/source staging, conversion); live-KB mutation stays serialized under the mutation lock, so raising it helps mainly when conversion is the bottleneck.
+
 `entity_types` (optional): a YAML list overriding the entity-type vocabulary used for entity pages; omit it to use the default `person`, `organization`, `place`, `product`, `work`, `event`, `other`.
 
 `extra_headers` (optional): a YAML mapping of extra HTTP headers sent with every LLM request (forwarded to LiteLLM's `extra_headers`). Useful for providers that expect custom headers, e.g. GitHub Copilot IDE-auth headers:
--- a/README.md
+++ b/README.md
@@ -333,6 +333,7 @@ OpenKB settings are initialized by `openkb init` and stored in `.openkb/config.y
 model: gpt-5.4                   # LLM model (any LiteLLM-supported provider)
 language: en                     # Wiki output language
 pageindex_threshold: 20          # PDF pages threshold for PageIndex
+file_processing_jobs: 2          # Files to prepare concurrently during `openkb add <dir>`
 ```
 
 Model names use `provider/model` LiteLLM [format](https://docs.litellm.ai/docs/providers) (OpenAI models can omit the prefix):
--- a/config.yaml.example
+++ b/config.yaml.example
@@ -2,6 +2,9 @@ model: gpt-5.4                   # LLM model (any LiteLLM-supported provider)
 language: en                     # Wiki output language
 pageindex_threshold: 20          # PDF pages threshold for PageIndex
 file_processing_jobs: 2          # Number of files to prepare concurrently during `openkb add <dir>`
+# Note: this parallelizes hashing/conversion/staging only. Live KB publish,
+# PageIndex indexing, LLM compilation, registry updates, and log writes remain
+# serialized under the KB mutation lock.
 
 # Optional: extra HTTP headers sent with every LLM request (forwarded to
 # LiteLLM's extra_headers). Some providers need these — e.g. GitHub Copilot
--- a/config.yaml.example
+++ b/config.yaml.example
@@ -1,6 +1,7 @@
 model: gpt-5.4                   # LLM model (any LiteLLM-supported provider)
 language: en                     # Wiki output language
 pageindex_threshold: 20          # PDF pages threshold for PageIndex
+file_processing_jobs: 2          # Number of files to prepare concurrently during `openkb add <dir>`
 
 # Optional: extra HTTP headers sent with every LLM request (forwarded to
 # LiteLLM's extra_headers). Some providers need these — e.g. GitHub Copilot
--- a/openkb/agent/compiler.py
+++ b/openkb/agent/compiler.py
@@ -31,6 +31,7 @@ import litellm
 from openkb import frontmatter
 from openkb.config import DEFAULT_ENTITY_TYPES, get_extra_headers, resolve_entity_types
 from openkb.lint import list_existing_wiki_targets, strip_ghost_wikilinks
+from openkb.locks import atomic_write_text
 from openkb.schema import INDEX_SEED, get_agents_md
 
 logger = logging.getLogger(__name__)
@@ -768,7 +769,7 @@ def _write_summary(wiki_dir: Path, doc_name: str, summary: str,
     fm_lines.append(f"doc_type: {doc_type}")
     fm_lines.append(_yaml_kv_line("full_text", f"sources/{doc_name}.{ext}"))
     fm_block = "---\n" + "\n".join(fm_lines) + "\n---\n\n"
-    (summaries_dir / f"{doc_name}.md").write_text(fm_block + summary, encoding="utf-8")
+    atomic_write_text(summaries_dir / f"{doc_name}.md", fm_block + summary)
 
 
 _SAFE_NAME_RE = re.compile(r'[^\w\-]')
@@ -828,7 +829,7 @@ def _write_concept(wiki_dir: Path, name: str, content: str, source_file: str, is
             if brief:
                 fm_lines.append(_yaml_kv_line("description", brief))
             existing = frontmatter.block(fm_lines) + clean
-            path.write_text(existing, encoding="utf-8")
+            atomic_write_text(path, existing)
             return
         # Guarantee type + refresh description on update; remove legacy brief:.
         ex_parts2 = frontmatter.split(existing)
@@ -840,7 +841,7 @@ def _write_concept(wiki_dir: Path, name: str, content: str, source_file: str, is
             # Drop legacy brief: lines (migrated to description:).
             fm_block = frontmatter.drop_line(fm_block, "brief")
             existing = fm_block + body
-        path.write_text(existing, encoding="utf-8")
+        atomic_write_text(path, existing)
     else:
         clean_parts = frontmatter.split(content)
         if clean_parts is not None:
@@ -852,7 +853,7 @@ def _write_concept(wiki_dir: Path, name: str, content: str, source_file: str, is
         if brief:
             fm_lines.append(_yaml_kv_line("description", brief))
         fm_block = "---\n" + "\n".join(fm_lines) + "\n---\n\n"
-        path.write_text(fm_block + content, encoding="utf-8")
+        atomic_write_text(path, fm_block + content)
 
 
 def _write_entity(
@@ -916,10 +917,10 @@ def _write_entity(
                     break
             merged = [source_file] + [s for s in recovered if s != source_file]
             existing = _build_entity_frontmatter(merged) + clean
-        path.write_text(existing, encoding="utf-8")
+        atomic_write_text(path, existing)
         return
 
-    path.write_text(_build_entity_frontmatter([source_file]) + clean, encoding="utf-8")
+    atomic_write_text(path, _build_entity_frontmatter([source_file]) + clean)
 
 
 _set_fm_line = frontmatter.set_line
@@ -1024,7 +1025,7 @@ def _add_related_link(
         text = _prepend_source_to_frontmatter(text, source_file)
 
     text += f"\n\nSee also: {link}"
-    path.write_text(text, encoding="utf-8")
+    atomic_write_text(path, text)
     return True
 
 
@@ -1051,7 +1052,7 @@ def _backlink_summary_pages(
     _ensure_h2_section(lines, section, quiet=True)
     for slug in reversed(missing):
         _insert_section_entry(lines, section, f"- [[{page_dir}/{slug}]]")
-    summary_path.write_text("\n".join(lines), encoding="utf-8")
+    atomic_write_text(summary_path, "\n".join(lines))
 
 
 def _backlink_pages(
@@ -1072,7 +1073,7 @@ def _backlink_pages(
         lines = text.split("\n")
         _ensure_h2_section(lines, "## Related Documents", quiet=True)
         _insert_section_entry(lines, "## Related Documents", f"- {link}")
-        path.write_text("\n".join(lines), encoding="utf-8")
+        atomic_write_text(path, "\n".join(lines))
 
 
 def _backlink_summary(wiki_dir: Path, doc_name: str, concept_slugs: list[str]) -> None:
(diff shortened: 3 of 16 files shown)


KylinMountain (commented):
Overall the design holds up well — staging isolation, serialized commit, the journal-based recovery, and the commit-time re-checks are all correct, and test coverage is solid. Nothing here blocks merge on day-to-day correctness. Three things are worth fixing first, though — #1 is the one that bites in normal use:

**1. Every per-file commit deep-copies the whole KB for the rollback backup** — `openkb/cli.py:475`, `openkb/mutation.py:165`

`_snapshot_add_paths` lists the entire `wiki/concepts` and `wiki/entities` trees plus the `.openkb/files` PDF blob store, and `snapshot_paths` `copytree`s them on *each* file's commit. A 50-file batch copies the whole KB ~50 times — O(files × corpus) I/O that scales with the KB and can fill the disk mid-batch; even adding a 2 KB `.md` copies the full blob store. Suggest snapshotting only the pages the doc actually touches (or hardlink/CoW), or taking the snapshot once per batch rather than once per file.

**2. Publish copies bytes instead of renaming, and drops the dir-fsync + chmod** — `openkb/mutation.py:18`

Staging is on the same filesystem as `raw/` and `wiki/sources/`, so `os.replace` would be an O(1) atomic rename instead of a full streaming copy + fsync. `_copy_file_atomic` also omits the parent-dir fsync and `fchmod` that `locks.atomic_write_bytes` performs, so published files are less durable across a crash and inherit `mkstemp`'s `0600` mode instead of the umask mode. Suggest `os.replace` with a copy fallback on `EXDEV`, reusing the `atomic_write_bytes` fsync/chmod path.

**3. Interrupted batches leak staging dirs** — `openkb/cli.py:1067`

Staging dirs are all `mkdir`'d up front in the futures comprehension, but `_cleanup_staging` only runs inside each per-file commit and there's no batch-level `try/finally`. A Ctrl+C, a failed file, or an mkdir error orphans the already-created `add-*` dirs, and recovery only scans `journal/` so it never reclaims them. Suggest a batch-level cleanup, or folding staging into the recovery sweep.

Happy to send a patch for #1 if useful. The rest of what I found is rarer crash/concurrency edges and design follow-ups — fine as separate issues.

gwokhou:
@KylinMountain  Thanks your review! All three items are addressed in the latest commits, with tests added for each change (full suite: 841 passed). I also found two further issues during self-review, noted below.

## Solved issues from your review https://github.com/VectifyAI/OpenKB/pull/104/changes/981e4918787dca158ef4329f96b8850f773a069a

### 1. Per-file snapshot cost reduced to O(touched).
wiki/concepts, wiki/entities, and .openkb/files are now snapshotted with hardlinks (os.link) rather than copytree, with EXDEV/EPERM/EACCES falling back to a copy. This removes the O(files × corpus) copy, including the case where a small add copied the entire blob store.

Notes:

* Hardlinks require every writer into these trees to write atomically (temp + replace, yielding a new inode). The wiki writers used in-place write_text, which would invalidate a hardlink backup on rewrite; all 14 have been switched to atomic_write_text. A regression test guards this — it fails if any writer reverts to in-place writes.
* Per-file rollback granularity is preserved. A single batch-level snapshot was not adopted because it would discard shared-index changes from earlier files when a later file fails.

One assumption worth confirming: .openkb/files relies on PageIndex appending only new {doc_id} blobs per add (verified in its local backend — fresh UUID per add, dedup returns early, no in-place rewrite). If you would prefer not to rely on that, the alternative is to track removals for new blobs in that tree.

Please share your perspective on the patch you had in mind. If a simpler approach than hardlinks plus atomic writers exists, I would prefer to adopt it.

### 2. Publish uses rename, durable, and umask-mode.
publish_staged_tree now uses os.replace (O(1)) with an EXDEV→copy fallback. The renamed file's data and parent directory are fsync'd, and the mode is the umask default rather than mkstemp's 0600, reusing locks._fsync_directory/_target_mode as suggested.

### 3. Interrupted batches reclaim their staging directories.
The batch records its staging directories and reclaims any that do not reach commit via a try/finally. It is scoped to the batch rather than folded into recovery, because the lock is released between per-file commits, and a recovery sweep could reclaim another in-flight batch's directories. Remaining limitation: a hard kill or power loss mid-batch can still orphan staging directories, since the final step does not run and recovery scans only journal/.

## Solved issues from self-review https://github.com/VectifyAI/OpenKB/pull/104/changes/5fb53b84187bc63113a72c0b07a6f52f437fbf00

* The publish rename path initially omitted the file-data fsync that the copy path performs, leaving committed metadata referencing undurable bytes after a crash. Fixed.
* The hardlink fallback handled only EXDEV/EPERM; on Windows, OneDrive/ACL hardlink blocks surface as EACCES and would abort the snapshot. EACCES is now included in the fallback set.

Additionally, recovery retry is now bounded (a persistently-failing rollback journal previously retried indefinitely; it now gives up after five attempts with an explicit warning), and rollback for hardlinked directories is O(touched) via inode comparison, leaving untouched files in place.

## Further Discussions

I am happy to hear the follow-up suggestions you mentioned (about rarer concurrency, crashes, and design).

gwokhou:
## One more solved issue from self-review bdef059

### a malformed recovery journal no longer bricks the KB lock

While self-reviewing this PR I caught a latent crash in `recover_pending_journals` (`openkb/mutation.py`).

#### Bug
The `except Exception` handler referenced `snapshot` unconditionally, but `snapshot` is assigned *inside* the `try` (after `json.loads` / `_snapshot_from_journal`). A malformed journal — a corrupt/empty/stray `.json` in `.openkb/journal/`, or a valid JSON missing the `kb_dir`/`backup_dir` keys — makes parsing/construction raise before the assignment, so the handler itself threw `UnboundLocalError`:

```python
except Exception as exc:
    snapshot.attempts += 1   # ← snapshot unbound → UnboundLocalError
```

Because journal draining runs on every **first exclusive** `kb_lock` acquisition (`locks._drain_pending_journals`), that propagated up *through* `kb_lock`: a single bad journal file would make every subsequent `add` / `remove` / `recompile` / `chat` fail with an opaque `UnboundLocalError`, bricking mutations for the entire KB.

Normal writes go through `atomic_write_json`, so the trigger for a malformed journal is uncommon (external corruption, manual edits, or a stray `.json` dropped in the dir) — but recovery must be robust to it regardless, and it wasn't.

#### Fix
Initialize `snapshot = None` per loop iteration, and in the `except` handle the `snapshot is None` case: best-effort `unlink` the unrecoverable journal, log loudly, and `continue` instead of crashing. The existing bounded-retry logic is left untouched for the case where `snapshot` *is* bound (a rollback/discard failure).

#### Test
Added a parametrized regression test (`tests/test_mutation.py`) over four malformed-journal payloads (empty, invalid JSON, missing keys, wrong shape), asserting that recovery returns cleanly and removes the bad journal — and, the key assertion, that the KB's mutation lock still acquires afterward. Full suite green (845 passed).

KylinMountain:
Re-verified all three against the actual code (including reading the PageIndex local backend), not just the writeup — they hold up:

- **#1** confirmed: hardlinked backups, O(touched) inode-diff rollback, and the EXDEV/EPERM/EACCES + non-persisted `hardlinked_dirs` fallbacks (safe full-copy on a crash-rebuilt snapshot) are all correct. I also confirmed PageIndex's local backend is genuinely append-only per add (fresh UUID, dedup short-circuits, the failure path unlinks only its own new blobs), so the `.openkb/files` hardlink assumption is sound.
- **#2** and **#3** look good; the hard-kill / power-loss staging residue is an acceptable, documented limit.

One thing I'd fold into *this* PR — cheap and on-theme:

**`lint.py:252` (`fix_broken_links`) still rewrites concept/entity pages in place** (`md.write_text(...)`). It's the lone exception to the "every writer into a hardlinked tree is atomic" invariant this PR introduces. It's safe *today* only because `fix_broken_links` is reachable solely from `remove` / `lint --fix`, neither of which snapshots. But the moment `hardlink_dirs` is reused for `remove`/`recompile` (the natural next step for crash-safety), this in-place write mutates the shared-inode backup and rollback silently becomes a no-op — and no current test catches it, since it's off the add path.

Two one-liners, both in scope here:
1. Switch that `write_text` → `atomic_write_text`. It's the 15th sibling of the 14 writers you already converted, and it makes "all wiki writes are atomic" a clean, assertable global property instead of "14 atomic + one in-place exception."
2. Make the invariant explicit in `snapshot_paths` (a cheap runtime assert or a docstring contract) so a future caller reusing `hardlink_dirs` can't step on it.

Actually wiring crash-safety into `remove`/`recompile` is genuinely separate work — good as a follow-up issue, not this PR.

For the rarer edges you asked about, I'll file these as separate issues:

- **SQLite hot-copy** — `pageindex.db` + `-wal`/`-shm` are snapshotted as independent byte copies; if a long-doc index leaves an uncheckpointed WAL, a rollback can restore a mutually-inconsistent db/wal pair. A checkpoint (or closing the client) before the snapshot would close it.
- **`convert_document` cross-thread re-entrancy** — lock reentrancy is thread-local, so calling the lock-acquiring `convert_document` from a prepare worker would deadlock. Latent (workers use `_convert_document_locked` today), guarded only by a docstring.
- **Advisory `doc_name` reservation** — the lock is released before commit, so a concurrent external `add` of a same-stem file makes this batch's file fail with the "conflict after parallel preparation" error instead of re-deriving a suffixed name.

Nice work overall — the self-review catches (rename data-fsync, EACCES fallback, bounded retry) were the right calls.

gwokhou:
@KylinMountain Thanks again for your quick and patient review! 

Issues fixed in 559ada7.

- `fix_broken_links` now uses `atomic_write_text`, so lint/remove cleanup no longer rewrites concept/entity pages in place.
- Added a regression test that runs `fix_broken_links` under a hardlinked `wiki/concepts` snapshot and verifies rollback restores the original wikilink content. This covers the off-add-path writer case you called out.
- Tightened the `snapshot_paths` docstring to make the hardlinked-dir writer contract explicit: writers must be atomic temp+replace or append-only.

Re-ran the focused coverage:

`UV_PYTHON=3.13 UV_CACHE_DIR=/tmp/uv-cache uv run --extra dev pytest tests/
  test_mutation.py tests/test_lint.py -q`

Result: `84 passed, 5 warnings`.

(Warnings are existing Python 3.13 SWIG/native-binding deprecation warnings. They are emitted during dependency import and are unrelated to the lint atomic-write / snapshot rollback change.)

I’m leaving remove/recompile crash-safety and the rarer SQLite/re-entrancy/doc-name reservation edges as follow-up issues, as suggested.

gwokhou:
Some issues found after 66765ff, working on it (already done)

gwokhou:
> Some issues found after [66765ff](https://github.com/VectifyAI/OpenKB/commit/66765ffef7b62b32864748543c140258a172ed91), working on it

Heads up on the commits after the merge — fixing gaps I caught reviewing the merged result. Resolving the upstream conflict pulled in the PageIndex Cloud import (#97), which predates this branch's journal layer, so cloud import was mutating the live KB with no snapshot/rollback — regressing the PR's own crash-safety contract.

- 5666951 wraps the `import_cloud_document → compile → registry` window in `snapshot_paths` (`mark_committed` after the registry write, rollback on any failure), makes post-commit `append_log` best-effort, and drops the now-redundant `_cleanup_failed_cloud_import`. Adds a regression test that `add --from-pageindex-cloud` drains a pending journal under the lock.

- 1be4350 fixes the first pass's whole-tree snapshot: snapshotting the whole `summaries`/`sources` trees was O(corpus) and could fail-before-work on a large KB. It splits `import_cloud_document` into `prepare_cloud_import` (read-only fetch + name resolution) so `doc_name` is known before writing, then snapshots only this doc's paths via the existing `_snapshot_add_paths`.

- eb4c1ab fixes a cross-system compatibility issue in that split: `prepare_cloud_import` no longer uses platform-native `Path(cloud_name).stem` for PageIndex Cloud display names. Cloud names are data, not local paths; Windows-style names such as `C:\Users\me\Cloud Paper.pdf` now resolve to the same `Cloud-Paper` doc_name on POSIX and Windows. Added a regression test so the POSIX-only wrong artifact (`C-Users-me-Cloud-Paper.json`) is not produced.

Verification:

Full suite verification for 5666951, 1be4350, eb4c1ab: **916 passed, 7 warnings in 4.88s**. `UV_CACHE_DIR=/tmp/uv-cache UV_PYTHON=3.13 uv run --extra dev pytest -q`
  
`import_cloud_document` is kept (thin `prepare` + write); `test_indexer.py` still exercises it. The separate Python 3.14 dependency metadata issue is project-level rather than this PR's cloud-import behavior, so I'm leaving that for a follow-up issue instead of widening this PR.

KylinMountain:
@gwokhou Honestly, this PR has gotten too big to review safely — +2600 across 16 files, and the upstream merge pulled cloud import in too. It's really two things bundled: crash-safe recovery (which is great, let's keep it) and concurrency.

And I'm not sure the concurrency earns its keep: only the prepare step runs in parallel — PageIndex + LLM compile, the actual slow parts, still run serially under the lock. So the speedup is small, but it's what forces all the tricky bits (name reservation, parallel conflict checks).

Can we split it? Land the serial recovery first — it's the valuable part and gets much simpler without concurrency. Do concurrency as a follow-up but maybe low priority? What do you think?

gwokhou:
> @gwokhou Honestly, this PR has gotten too big to review safely — +2600 across 16 files, and the upstream merge pulled cloud import in too. It's really two things bundled: crash-safe recovery (which is great, let's keep it) and concurrency.
> 
> And I'm not sure the concurrency earns its keep: only the prepare step runs in parallel — PageIndex + LLM compile, the actual slow parts, still run serially under the lock. So the speedup is small, but it's what forces all the tricky bits (name reservation, parallel conflict checks).
> 
> Can we split it? Land the serial recovery first — it's the valuable part and gets much simpler without concurrency. Do concurrency as a follow-up but maybe low priority? What do you think?

I agree. I'll close this PR first, and then split it into two PRs: 
1. Landing the serial crash-safe recovery/journal rollback work, keeping the cloud-import recovery fixes only where they preserve that same contract. 
2. Moving the concurrency / parallel prepare path into a follow-up PR and treating it as a lower priority unless we can demonstrate that the speedup justifies the added complexity.

PR #120, opened 2026-06-20
fix(skill): resolve references/ wikilinks whose stem contains a dot

jichaowang02-lang:
## Summary

`validate_skill` reports a false **"doesn't exist"** error for a valid
`[[references/...]]` wikilink whose target stem contains a dot (e.g.
`[[references/api.v2]]`, `[[references/v1.2-guide]]`), even when the file
(`references/api.v2.md`) is present and correctly linked.

## Root cause

`openkb/skill/validator.py` only appended the implicit `.md` when
`Path(link).suffix` was empty:

```python
target = refs_dir / link
if not target.suffix:            # <-- ".v2" is a truthy suffix
    target = target.with_suffix(".md")
if not target.exists():          # looks for "references/api.v2" (no .md) -> false error
```

`Path.suffix` returns everything after the **last** dot, so `api.v2` has a
truthy suffix (`.v2`). The `.md` is never appended, and the validator then
checks for an extension-less `references/api.v2`, which doesn't exist —
producing a false error. `WIKILINK_RE` (`[a-z0-9._/-]+`) allows `.` in the
target, so such links are legal and reach this branch.

| link | file on disk | before | after |
|---|---|---|---|
| `[[references/topic]]` | `topic.md` | ✅ resolves | ✅ (unchanged) |
| `[[references/api.v2.md]]` | `api.v2.md` | ✅ resolves | ✅ (unchanged) |
| `[[references/api.v2]]` | `api.v2.md` | ❌ false "doesn't exist" | ✅ resolves |
| `[[references/ghost]]` | *(absent)* | ✅ errors | ✅ errors (unchanged) |

## Fix

Append `.md` based on a literal `.md` check instead of `Path.suffix`:

```python
target = refs_dir / (link if link.lower().endswith(".md") else f"{link}.md")
```

A dotted stem is now suffixed correctly; an explicit `...md` link is left
alone; genuinely missing references still error.

## Testing

```text
$ pytest tests/test_skill_validator.py -q
30 passed
$ ruff check openkb/skill/validator.py tests/test_skill_validator.py
All checks passed!
```

Adds `test_wikilink_dotted_stem_without_md_suffix_resolves` — the existing
`test_wikilink_without_md_suffix_resolves` only covered a dot-free stem, so
this path was uncovered.

### DIFF
--- a/openkb/skill/validator.py
+++ b/openkb/skill/validator.py
@@ -212,6 +212,15 @@ def validate_skill(skill_dir: Path, *, strict: bool = False) -> ValidationResult
         # "api.v2" has a truthy suffix (".v2"), so Path.suffix would skip the
         # ".md" and then look for a non-existent extension-less file.
         target = refs_dir / (link if link.lower().endswith(".md") else f"{link}.md")
+        # A reference must resolve to a file *under* references/ — reject any
+        # that escape it (e.g. "[[references/../SKILL]]"), which would also
+        # break the relative_to(skill_dir) call in the not-found message below.
+        if not target.resolve().is_relative_to(refs_dir.resolve()):
+            result.errors.append(
+                f"SKILL.md references [[references/{link}]] which resolves "
+                f"outside the references/ directory."
+            )
+            continue
         if not target.exists():
             result.errors.append(
                 f"SKILL.md references [[references/{link}]] but "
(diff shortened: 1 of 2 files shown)


KylinMountain (approved):
Thanks @jichaowang02-lang — clean fix. The dotted-stem resolution is correct, and the escape guard is a nice catch (the old code silently accepted [[references/../SKILL]]). LGTM 👍

PR #121, opened 2026-06-20
fix(compiler): preserve frontmatter when closing "---" has no trailing newline

jichaowang02-lang:
## Summary

A frontmatter-only concept/entity page that ends right at the closing `---`
with **no trailing newline** gets its frontmatter silently destroyed when a
source is added or removed: the opening `---`, the `type` / `description`
keys, and the existing `sources:` list are all dropped — corrupting the page
and losing source provenance.

## Root cause

`frontmatter.split` is lossless and, for a page with no trailing newline after
the closing delimiter, returns a block ending in a bare `\n---` (not
`\n---\n`):

```python
# openkb/frontmatter.py — split()
after = text.find("\n", nl + 1)   # newline ending the closing '---' line
if after == -1:
    return text, ""               # closing '---' has no trailing newline
```

But both `_prepend_source_to_frontmatter` and `_remove_source_from_frontmatter`
located the closing delimiter with `fm_block.rpartition("\n---\n")`, which
matches nothing in that case — so `fm_prefix` becomes `""`, every frontmatter
line is lost, and the block is rebuilt from nothing:

```python
fm_prefix, _, _ = fm_block.rpartition("\n---\n")   # -> ("", "", fm_block)
fm_lines = fm_prefix.split("\n")                   # -> [""]
```

Concrete trace (`_prepend_source_to_frontmatter`):

```
in : '---\nsources: ["summaries/p1.md"]\ntype: "Concept"\ndescription: "Focus"\n---'
out: '\nsources: ["summaries/p2.md"]\n---\n'     # type, description, p1.md, opening --- all gone
```

This is reached during normal use via `_add_related_link` (run for every
related concept/entity at compile time) and the `openkb remove` flow
(`remove_doc_from_{concept,entity}_pages`).

## Fix

Strip whichever closing form is actually present and re-append the same one:

```python
closing = "\n---\n" if fm_block.endswith("\n---\n") else "\n---"
fm_prefix = fm_block[: -len(closing)]
```

With the fix the same input yields
`---\nsources: ["summaries/p2.md", "summaries/p1.md"]\ntype: "Concept"\ndescription: "Focus"\n---`
— all keys and the prior source preserved. Pages that have a body (the common
case) are byte-for-byte unaffected.

## Testing

```text
$ pytest tests/test_compiler.py::TestFrontmatterSourceMutation -q
3 passed
$ ruff check openkb/agent/compiler.py
All checks passed!
```

Adds `TestFrontmatterSourceMutation` (prepend + remove on a no-trailing-newline
page, plus a with-body regression guard). The existing `_add_related_link` /
`remove_doc_from_entity_pages` tests only used pages **with** a body, so this
path was uncovered.

> Note: the rest of `tests/test_compiler.py` has some pre-existing failures
> in the LLM-mock suites that are unrelated to this change (they reproduce on
> `main` without this patch); the frontmatter tests above pass.

### DIFF
--- a/openkb/agent/compiler.py
+++ b/openkb/agent/compiler.py
@@ -940,12 +940,15 @@ def _prepend_source_to_frontmatter(text: str, source_file: str) -> str:
         return text
 
     fm_block, body = parts
-    # Split the fm_block into lines for per-line manipulation. fm_block ends
-    # with "\n---\n"; strip the trailing closing delimiter + newline to get
-    # the prefix lines (opening "---" + content lines), then re-append after.
-    fm_prefix, _, _ = fm_block.rpartition("\n---\n")
+    # Strip the trailing closing delimiter to get the prefix lines (opening
+    # "---" + content lines), then re-append it. `frontmatter.split` leaves the
+    # closing at the end of fm_block as either "\n---\n" or a bare "\n---" (when
+    # the page ends at the delimiter with no trailing newline). Assuming only
+    # "\n---\n" would, for the bare form, make the strip below collapse the
+    # whole block and drop every existing frontmatter key.
+    closing = "\n---\n" if fm_block.endswith("\n---\n") else "\n---"
+    fm_prefix = fm_block[: -len(closing)]
     fm_lines = fm_prefix.split("\n")
-    closing = "\n---\n"
 
     for i, line in enumerate(fm_lines):
         if not line.lstrip().startswith("sources:"):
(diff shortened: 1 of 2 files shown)


KylinMountain (approved):
Thanks @jichaowang02-lang — nice catch. The no-trailing-newline corruption is a real one (hits via _add_related_link and openkb remove), and the fix handles both closing forms cleanly. LGTM 👍

PR #122, opened 2026-06-20
fix(images): don't overwrite relative images that share a basename

jichaowang02-lang:
## Summary

`copy_relative_images` could **silently lose an image** and render the wrong
one. It named every destination after `src.name` (the basename only), so two
references to *different* files that happen to share a basename collide.

## Root cause

```python
filename = src.name          # basename only
dest = images_dir / filename
shutil.copy2(src, dest)      # second same-basename source overwrites the first
```

For `a` + `b`:
- both copy to `images_dir/logo.png` → the second `copy2` overwrites the first
  (image A's bytes are gone), and
- both links are rewritten to `sources/images/<doc>/logo.png`, so the rendered
  document shows the **same** image in both places.

`extract_base64_images` already avoids this by numbering destinations
(`img_{counter:03d}`); only the relative-image path was affected.

| input | before | after |
|---|---|---|
| `a/logo.png` + `b/logo.png` (distinct) | 1 file, both links → it, A lost | 2 files (`logo.png`, `logo_1.png`), distinct links |
| `logo.png` referenced twice (same file) | 1 file | 1 file (deduped, both links agree) |

## Fix

Track the destination assigned to each source: reuse it when the **identical**
source is referenced more than once (no duplicate copy), and disambiguate with
a `{stem}_{n}{suffix}` suffix when a **different** source would collide on a
name already taken.

## Testing

```text
$ pytest tests/test_images.py -q
12 passed
$ ruff check openkb/images.py tests/test_images.py
All checks passed!
```

Adds `test_same_basename_different_dirs_no_overwrite` (both images preserved,
links distinct) and `test_same_image_referenced_twice_is_copied_once` (dedup).

### DIFF
--- a/openkb/images.py
+++ b/openkb/images.py
@@ -229,7 +229,12 @@ def copy_relative_images(
     # ``b/logo.png``) don't overwrite each other and collapse both links onto a
     # single image.
     assigned: dict[Path, str] = {}
-    taken: set[str] = set()
+    # Seed taken names from any files already in images_dir so a re-convert into
+    # an existing per-doc directory disambiguates against them instead of
+    # overwriting a previously-copied image with a same-basename source.
+    taken: set[str] = (
+        {p.name for p in images_dir.iterdir()} if images_dir.is_dir() else set()
+    )
 
     for match in _RELATIVE_RE.finditer(markdown):
         alt, rel_path = match.group(1), match.group(2)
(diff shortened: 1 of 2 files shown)


KylinMountain (commented):
Core fix (first commit) is correct.

The second commit — seeding `taken` from `images_dir`, per Copilot's note — is the problem. This dir is per-`doc_name`, so it only ever holds this doc's own images from a prior run; overwriting them on re-convert is the desired refresh, not data loss. The real bug (two *different* sources colliding in one pass) is already prevented by `taken` starting empty. Seeding from disk instead makes re-convert non-idempotent — stale orphans + `_1`/`_2` link churn on every edit.

Suggest dropping the second commit — the first one alone is the right fix.

PR #142, opened 2026-06-26
feat: add serial crash-safe mutation recovery

gwokhou:
## TL;DR

This PR makes document import fail more cleanly. If `openkb add` is interrupted or errors halfway through, OpenKB should no longer leave half-written KB files behind; it can recover the previous stable state before accepting the next mutation.

## Summary

Adds crash-safe mutation recovery for serial KB mutations without bringing back the concurrent add pipeline from #104.

This PR keeps the scope focused on the first split-out piece from the abandoned PR: journaled snapshot/rollback for `add` and PageIndex Cloud import, plus the atomic writer prerequisites needed for hardlink-backed rollback.

## What changed

- Added `openkb/mutation.py` with:
  - mutation snapshots
  - active/committed/rolled_back journals
  - rollback and recovery of interrupted mutations
  - bounded rollback retry
  - malformed journal handling
  - staged artifact publishing with atomic rename + EXDEV copy fallback

- Wired recovery into serial mutation paths:
  - `openkb add <file>` converts into staging, snapshots final KB paths, publishes, compiles, then commits via registry write + journal mark
  - `openkb add --from-pageindex-cloud` prepares cloud data read-only, snapshots doc-specific paths, writes/compiles/registers under the same recovery contract
  - exclusive KB lock acquisition drains pending journals before mutating

- Kept conversion serial while adding staging support:
  - `convert_document(..., staging_dir=...)` writes raw/source artifacts into isolated staging before commit
  - no parallel prepare hook or doc-name override path is introduced

- Made wiki writers atomic where hardlink snapshots depend on temp+replace semantics:
  - compiler summary/concept/entity/index writes
  - lint broken-link cleanup

## Explicitly out of scope

This does not reintroduce the concurrency half of #104:

- no `ThreadPoolExecutor`
- no `file_processing_jobs`
- no jobs>1 directory add path
- no parallel prepare pipeline
- no batch doc-name reservation logic

## Validation

- `UV_CACHE_DIR=/tmp/uv-cache UV_PYTHON=3.13 uv run --extra dev pytest -q`
  - `898 passed, 5 warnings`

- `UV_CACHE_DIR=/tmp/uv-cache UV_PYTHON=3.13 uv run --extra dev pytest tests/test_add_command.py::TestImportFromPageindexCloud tests/test_mutation.py -q`
  - `25 passed`

- `UV_CACHE_DIR=/tmp/uv-cache UV_PYTHON=3.13 uv run --extra dev pytest tests/test_converter.py tests/test_add_command.py tests/test_mutation.py -q`
  - `72 passed`

- `UV_CACHE_DIR=/tmp/uv-cache UV_PYTHON=3.13 uv run --extra dev ruff check openkb/converter.py openkb/mutation.py tests/test_mutation.py`
  - passed

## Notes

Full-project `ruff` and `ty` are not clean on current `main`; this PR fixes the new diagnostics introduced by the recovery changes but does not widen scope to existing lint/type debt.

### DIFF
--- a/openkb/agent/compiler.py
+++ b/openkb/agent/compiler.py
@@ -36,6 +36,7 @@ from openkb.config import (
     resolve_entity_types,
 )
 from openkb.lint import list_existing_wiki_targets, strip_ghost_wikilinks
+from openkb.locks import atomic_write_text
 from openkb.schema import INDEX_SEED, get_agents_md
 
 logger = logging.getLogger(__name__)
@@ -779,7 +780,7 @@ def _write_summary(wiki_dir: Path, doc_name: str, summary: str,
     fm_lines.append(f"doc_type: {doc_type}")
     fm_lines.append(_yaml_kv_line("full_text", f"sources/{doc_name}.{ext}"))
     fm_block = "---\n" + "\n".join(fm_lines) + "\n---\n\n"
-    (summaries_dir / f"{doc_name}.md").write_text(fm_block + summary, encoding="utf-8")
+    atomic_write_text(summaries_dir / f"{doc_name}.md", fm_block + summary)
 
 
 _SAFE_NAME_RE = re.compile(r'[^\w\-]')
@@ -839,7 +840,7 @@ def _write_concept(wiki_dir: Path, name: str, content: str, source_file: str, is
             if brief:
                 fm_lines.append(_yaml_kv_line("description", brief))
             existing = frontmatter.block(fm_lines) + clean
-            path.write_text(existing, encoding="utf-8")
+            atomic_write_text(path, existing)
             return
         # Guarantee type + refresh description on update; remove legacy brief:.
         ex_parts2 = frontmatter.split(existing)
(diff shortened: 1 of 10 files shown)


KylinMountain:
Solid PR — the snapshot/rollback design is sound and the hardlink backup contract holds (every writer into the hardlinked dirs is atomic temp+replace or append-only).

One gap worth closing before merge, since it's the PR's own goal: **the URL path isn't crash-safe.** `openkb add <url>` calls `add_single_file(..., stage=False)` (cli.py:964), so `staging_dir=None` and `convert_document` (cli.py:444) writes into the live KB *before* `snapshot_paths()` (cli.py:461). A crash mid-conversion leaves a partial `wiki/sources/<doc>.md` with no journal to roll back — exactly the orphan state the PR eliminates for the local/cloud paths. Could the URL path use `stage=True` and handle the dedup cleanup separately?

gwokhou:
> Solid PR — the snapshot/rollback design is sound and the hardlink backup contract holds (every writer into the hardlinked dirs is atomic temp+replace or append-only).
> 
> One gap worth closing before merge, since it's the PR's own goal: **the URL path isn't crash-safe.** `openkb add <url>` calls `add_single_file(..., stage=False)` (cli.py:964), so `staging_dir=None` and `convert_document` (cli.py:444) writes into the live KB _before_ `snapshot_paths()` (cli.py:461). A crash mid-conversion leaves a partial `wiki/sources/<doc>.md` with no journal to roll back — exactly the orphan state the PR eliminates for the local/cloud paths. Could the URL path use `stage=True` and handle the dedup cleanup separately?

@KylinMountain Thanks for the CR, gap fixed🫡

KylinMountain (approved):
LGTM

PR #147, opened 2026-06-27
fix(lint): don't hide same-stem orphans behind qualified links

jichaowang02-lang:
## Summary

`find_orphans` builds the set of linked pages by adding, for **every** wikilink, both its full path and its bare **stem**:

```python
for lnk in links:
    incoming.add(lnk.strip().strip("/"))
for lnk in links:
    incoming.add(Path(lnk.strip()).stem)   # stem of EVERY link, even qualified ones
```

So a qualified link like `[[concepts/foo]]` also registers the bare stem `foo`. Any other page that merely shares that stem in a different directory — e.g. `summaries/foo` — is then treated as "linked" (`stem in incoming`) and excluded from orphan detection. **A genuine orphan is silently missed.**

## Fix

Only register the bare stem when the link itself is a bare stem (no `/`). Qualified links register only their full path:

```python
cleaned = lnk.strip().strip("/")
incoming.add(cleaned)
if "/" not in cleaned:
    incoming.add(Path(cleaned).stem)
```

Bare `[[foo]]` links still resolve by stem (a page named `foo` in any directory stays non-orphaned), so the intended bare-link behaviour is preserved.

## Testing

```text
$ pytest tests/test_lint.py::TestFindOrphans -q
6 passed
```

- `test_qualified_link_does_not_hide_same_stem_orphan` — `summaries/dup` is now flagged when only `[[concepts/dup]]` is linked (**fails before the fix**).
- `test_bare_stem_link_still_matches_same_stem_page` — a bare `[[dup]]` link still marks a same-stem page as non-orphan.

(The 4 unrelated `TestFindMissingEntriesRegistry` failures are pre-existing on `main` and untouched by this change.)

### DIFF
--- a/openkb/lint.py
+++ b/openkb/lint.py
@@ -321,10 +321,14 @@ def find_orphans(wiki: Path) -> list[str]:
     incoming: set[str] = set()
     for links in outgoing.values():
         for lnk in links:
-            incoming.add(lnk.strip().strip("/"))
-        # Also add stems
-        for lnk in links:
-            incoming.add(Path(lnk.strip()).stem)
+            cleaned = lnk.strip().strip("/")
+            incoming.add(cleaned)
+            # Only treat a link as a bare-stem reference when it is itself a
+            # bare stem. A qualified link like ``concepts/foo`` must not mark a
+            # same-stem page in another directory (``summaries/foo``) as linked,
+            # which would hide a genuine orphan.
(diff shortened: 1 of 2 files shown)


KylinMountain (approved):
LGTM. Thanks for the sharp catch and the clean fix!  Much appreciated 🙏

Issue #37, opened 2026-05-04
feat: add cache_control breakpoints to compiler for Anthropic prompt caching


pitimon:
## Problem

OpenKB compiler reuses a "base context A" (system + document) across N+M+2 LLM calls per document (summary → concepts-plan → N create + M update concept pages). Without `cache_control` markers, every call re-bills the full document content as input tokens.

For Anthropic Sonnet 4.5 (via OpenRouter or direct), prompt caching can reduce input cost by ~90% on cached prefix and reduce TTFT. Minimum cacheable size is 1,024 tokens, easily exceeded by typical document content.

## Proposal

Add `cache_control: {"type": "ephemeral"}` markers at two breakpoints in `openkb/agent/compiler.py`:

1. **End of `doc_msg`** in `compile_short_doc` + `compile_long_doc` — caches `system + doc` for all downstream calls (summary, plan, every concept).
2. **End of assistant summary message** in `_compile_concepts` (3 call sites: plan, create, update) — caches `system + doc + summary` for all concept generation calls.

Two breakpoints, well within Anthropic's max-4 limit.

## Compatibility

- Anthropic / OpenRouter→Anthropic: cache_control honored.
- OpenAI: list-of-blocks content format is valid (Vision API uses it); cache_control silently ignored.
- Other providers: LiteLLM normalizes/strips unknown fields.

## Side fix

`_llm_call_async` currently does not forward `**kwargs` while `_llm_call` does (asymmetry noted in memory #82886). Add `**kwargs` for parity.

## Out of scope

- OpenRouter Response Caching (`X-OpenRouter-Cache: true`) — different mechanism, evaluated separately.
- Refactoring messages into a dedicated builder module — keep patch surgical.

## Test plan

- Existing pytest suite passes (mocks accept `*args, **kwargs`).
- New assertion: completion payload contains `cache_control` block on the doc_msg.
- Manual smoke against a real Anthropic key: observe `cached_tokens` in `prompt_tokens_details` on calls 2..N.

## References

- Memory observation S11144 (3-5 line patch feasibility, OpenKB compiler audit).
- CLAUDE.md compiler architecture: "Designed around prompt-cache reuse: a single base context A reused across summary → concept-plan → concept-page calls."

KylinMountain:
thanks for your feedback and fix,  since it is fixed, I will close this issue.

PR #154, opened 2026-06-30
fix(compiler): strip cache_control for non-Anthropic providers

Aldominguez12:
## Problem

`_cached_text` tags reusable prompt context (system prompt, document, summary, known-targets) with an Anthropic ephemeral `cache_control` marker — added in #37 for Anthropic prompt caching. Its docstring assumes non-supporting providers simply *ignore* the marker:

> For providers that ignore cache_control, the list-of-blocks payload remains a valid OpenAI-compatible content shape.

That assumption does not hold for **Gemini**. LiteLLM translates the `cache_control` block into a Gemini `CachedContent` object, which then conflicts with `system_instruction`/`tools`, so every request fails with:

```
400 CachedContent can not be used with system_instruction, tools or tool_config ...
```

Net effect: **every compile with a `gemini/*` model fails out of the box** — anything that routes through the compiler (`openkb add`, `lint`, `query`, `skill`, `deck`). Reproduced on `gemini/gemini-2.5-pro` and `gemini/gemini-3-flash-preview` with v0.4.2.

## Fix

Strip the marker at the single request egress (`_llm_call` / `_llm_call_async`) for any provider that won't honour it, keeping it for Anthropic (direct) and Claude served via OpenRouter / Bedrock / Vertex. Producers keep tagging optimistically; only the egress decides whether the marker survives.

- Anthropic prompt caching is unchanged.
- Gemini's *implicit* caching still applies to the remaining plain-text blocks (verified the `cached=` token count is still reported).
- `supports_prompt_caching()` is deliberately **not** used as the gate: it returns `True` for Gemini and GPT-4o (they have *some* caching), which is not the same as accepting Anthropic's `cache_control` block. Provider identity via `get_llm_provider` is the correct signal.

## Tests

`TestCacheControlStripping` (5 tests): provider gating (Anthropic / OpenRouter-Anthropic kept; Gemini / GPT stripped), non-mutating marker removal, and the keep/strip paths through `_llm_call`. The existing `TestCacheControl` (Anthropic still gets breakpoints) and `TestLLMCallExtraHeaders` continue to pass.

> `get_llm_provider` is imported locally inside the gate so provider detection stays correct even under tests that patch the module-level `litellm`.

### DIFF
--- a/openkb/agent/compiler.py
+++ b/openkb/agent/compiler.py
@@ -262,10 +262,73 @@ def _cached_text(text: str) -> list[dict]:
     ephemeral cache_control marker.
 
     LiteLLM passes the marker through to Anthropic (and OpenRouter →
-    Anthropic). For providers that ignore cache_control, the list-of-blocks
-    payload remains a valid OpenAI-compatible content shape.
+    Anthropic). For other providers the marker is stripped at the request
+    egress (see :func:`_strip_cache_control`, applied in :func:`_llm_call`),
+    because not every provider merely *ignores* it — Gemini in particular
+    turns it into a 400. The list-of-blocks payload that remains is a valid
+    OpenAI-compatible content shape.
     """
     return [{"type": "text", "text": text, "cache_control": {"type": "ephemeral"}}]
+
+
+def _accepts_cache_control(model: str) -> bool:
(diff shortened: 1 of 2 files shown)


KylinMountain (approved):
LGTM, thanks for the thorough fix.

PR #163, opened 2026-07-02
fix(converter): preserve docx embedded images from markitdown

### LINKED ISSUE #162: DOCX embedded images are lost during conversion
OpenKB currently converts `.docx` files via `MarkItDown()` and then runs `extract_base64_images(...)` on the generated markdown.

However, for `.docx` files with embedded images, MarkItDown needs `keep_data_uris` enabled to preserve full base64 image content. Otherwise the output may contain incomplete image refs like ``, which means OpenKB cannot extract and persist those images afterward.

Relevant upstream context:
- https://github.com/microsoft/markitdown/issues/1222

Suggested fix:
- enable `keep_data_uris` in the MarkItDown-based `.docx` conversion path before calling `extract_base64_images(...)`

RussellLuo:
Fixes #162.

### DIFF
--- a/openkb/converter.py
+++ b/openkb/converter.py
@@ -230,7 +230,7 @@ def convert_document(
             markdown = convert_pdf_with_images(src, doc_name, images_dir)
         else:
             # Non-PDF, non-MD: use markitdown (docx, pptx, html, etc.)
-            mid = MarkItDown()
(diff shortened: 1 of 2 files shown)


quanqigu (approved):
Great catch, appreciate that. LGTM

PR #165, opened 2026-07-02
fix(converter): move keep_data_uris to convert call and fix lint

RussellLuo:
Sorry about that. In PR #163, the code I verified locally did not match what I actually committed, and I also had not run the test locally. This follow-up fixes the incorrect change and the leftover lint issue.


KylinMountain:
Thanks for the fix, @RussellLuo — I verified it end-to-end (a PPTX with an embedded image now yields img_001.png where 0.4.3 dropped it), and it's green across ruff/mypy/pytest. Merging now. 🙏

KylinMountain (approved):
LGTM

PR #156, opened 2026-07-01
feat(mutation): route serial add mutations through coordinator

gwokhou:
## Background

#142 made serial `openkb add` crash-safe with journals, touched-path snapshots, staged publish, rollback, and recovery drain. The recovery model is now in place, but its orchestration still lives inline in the CLI add paths.

**Before OpenKB can safely add prepare workers or `--jobs`, official KB mutation ownership needs to be explicit.** Future workers may prepare disposable artifacts, but one serial owner must remain responsible for snapshotting rollback-critical paths, publishing staged artifacts, updating the registry, marking the commit point, and handling rollback or post-commit cleanup.

This PR introduces that owner for add-like mutations. The coordinator lifts the crash-safe lifecycle from #142 into a higher-level abstraction: it owns transaction mechanics, while the add paths keep document-specific publish, compile, and register logic.

This is the Stage 3 bridge from the parallel architecture roadmap: https://github.com/VectifyAI/OpenKB/discussions/151. It keeps ingestion serial, makes the mutation boundary reviewable, and gives future non-add commands a pattern to adopt when they need the same recovery guarantees.

## Summary

- add `AddMutationPlan` and `run_add_mutation` as the serial coordinator for add-like KB mutations
- route local file add through the coordinator
- route URL add through the coordinator after the URL is fetched into `raw/`
- route PageIndex Cloud import through the same coordinator
- make touched paths, mutation body, commit point, rollback behavior, staging cleanup, and post-commit hooks explicit
- preserve the crash-safe recovery behavior introduced in #142
- keep the `.openkb/files` rollback surface intact without eager snapshotting — the long-doc blob is registered lazily via `snapshot.track_new()`, and `index_long_document` deletes the PageIndex doc if it fails after `col.add()` so a half-applied add leaves no orphan blob (review-driven fix for a gap left by #155)

## Scope

This is the Stage 3 slice from the parallel architecture roadmap: https://github.com/VectifyAI/OpenKB/discussions/151

In scope:
- existing `openkb add <file>`
- existing `openkb add <URL>` after URL fetch
- existing `openkb add --from-pageindex-cloud <DOC_ID>`
- add mutation snapshot/rollback orchestration
- post-commit ingest log hooks
- tests for coordinator routing, rollback behavior, URL add regression coverage, and interrupted cloud import rollback

Out of scope:
- coordinator ownership of URL fetching itself
- `openkb add <dir> --jobs N`
- prepare/commit split
- worker pools
- deterministic parallel commit queues
[...]

PR #172, opened 2026-07-04
refactor(mutation): split lock-free prepare from serial commit for directory ingest

gwokhou:
## Background

#142 made serial `openkb add` crash-safe (journals, touched-path snapshots, staged publish, rollback, recovery drain), and #156 gave those mechanics an explicit owner — `AddMutationPlan` / `run_add_mutation` — so one serial path is responsible for every mutation of official KB state, while anything else may only write disposable staging.

That split is what makes parallel ingest safe to build. But there is no real "prepare" step yet: `add` still converts and commits in a single pass under the lock. **Before `--jobs` can run conversions in parallel, prepare has to become a lock-free phase of its own — one that writes only private staging and never touches the registry or official `raw/`/`wiki/`.** The serial owner then commits under the lock, resolving the final name and publishing, just as it does today.

This PR adds that prepare/commit split for `openkb add <dir>`. Ingestion stays serial for now — prepare still runs one file at a time under the single held `kb_ingest_lock` — but it no longer needs the lock or touches official state, which is exactly the property Stage 5 will parallelize.

This is the Stage 4 slice from the parallel architecture roadmap: https://github.com/VectifyAI/OpenKB/discussions/151.

## Summary

No user-visible behavior change: `openkb add <dir>` produces identical results — this only factors conversion into a lock-free prepare + serial commit so Stage 5 can parallelize prepare.

- add `convert_document_for_prepare`: lock-free conversion into private `.openkb/staging/prepare/` under a placeholder `doc_name` (sanitized stem); returns `ConvertResult` without registering the hash or resolving the final name
- add `prepare_document` owning the staging-dir lifecycle (`rmtree` on interrupt); prepare/commit are coordinator-internal, called only by `add_directory_serial`
- add `add_directory_serial`: a serial prepare → commit loop for `openkb add <dir>`, run under the `add` command's existing `@_with_kb_lock`
- add `commit_prepared_document`: requires `kb_ingest_lock` held (reentrant acquire), then commits via the prepared branch of `_add_single_file_locked`
- re-validate under the lock at commit because prepare ran without it: re-decide skip from live registry state, re-hash the source, and re-convert when the source changed or prepare short-circuited with no artifacts (the stale-prepare contract)
- retarget staged raw/source/images from the placeholder name to the owner-resolved final name before publish
- add `_reap_prepare_staging`: reclaim orphaned prepare staging at first exclusive-lock acquisition; skip symlinks, unlink stray files, no per-prepare marker needed

## Scope

In scope:
- existing `openkb add <dir>`
- lock-free prepare into private staging
- serial commit under the existing coordinator owner
- prepare-staging reaper
- tests for prepare (private staging, no lock), commit (final-name resolution, lock required), reaper (orphans + symlinks), stale-prepare and source-changed TOCTOU re-prepare, end-to-end `add <dir>`, and per-file failure isolation

Out of scope:
- `openkb add <dir> --jobs N` (Stage 5)
- worker pools / parallel prepare
- deterministic parallel commit queues
- routing single-file add, URL add, or PageIndex Cloud import through the prepare/commit split (they keep using `convert_document` directly under the coordinator)

## Why this matters

After this PR, the prepare phase is a self-contained, lock-free, worker-safe unit that writes only disposable staging. Stage 5 can then add `--jobs` by running prepares in parallel and feeding `PreparedDocument`s into the unchanged serial commit owner — final-name resolution, registry writes, commit signals, rollback, and cleanup all stay under the single owner established in #156.

## Verification

- `UV_PYTHON=3.13 UV_CACHE_DIR=/tmp/uv-cache uv run --extra dev pytest tests/test_add_prepare.py tests/test_add_command.py tests/test_converter.py tests/test_locks.py -q`
- `UV_PYTHON=3.13 UV_CACHE_DIR=/tmp/uv-cache uv run --extra dev mypy openkb`
- `git diff --check main...HEAD`


gwokhou:
### `6969833` — Windows hardening for the lock-free prepare path

Follow-up to `ee22914`. That commit introduced the prepare/commit split (`_reap_prepare_staging`, `_retarget_prepared_document_artifacts`) with three latent Windows-only defects. Landed as a separate commit rather than force-pushing an amend into the already-pushed `ee22914`.

**1. Reaper no longer stalls the whole KB on a locked loose file.**
`_reap_prepare_staging`'s file branch used `unlink(missing_ok=True)`, which only swallows `FileNotFoundError`. On Windows, an AV/indexer holding a loose file in `staging/prepare/` makes `unlink` raise `PermissionError`, which escaped `kb_lock` and crashed every exclusive-lock command (`add`/`remove`/`recompile`/`chat`) until the handle released. The directory branch was already safe (`rmtree(ignore_errors=True)`); the file branch is now wrapped in `try/except OSError` to match. *(Trigger is narrow — staging normally holds dirs, not loose files — but the consequence was a full KB stall.)*

**2. Read-only staging now self-heals.** `shutil.copy2` preserves a read-only source's attribute into staging; on Windows `rmtree(ignore_errors=True)` can't delete read-only entries, so an orphaned staging dir re-logged *"Could not fully reap…"* on every lock acquisition and never cleared. `rmtree` now passes an `onerror` that clears the read-only bit and retries. POSIX is unaffected (a read-only bit never blocks `unlink` there).

**3. Retarget preserves LF.** `_retarget_prepared_document_artifacts` used `Path.write_text` without `newline=`, so on Windows `\n`→`\r\n` and a collision-renamed source became CRLF while every other source (written via `atomic_write_text`) stayed LF. Switched to `atomic_write_text`.

**Testing:** CI is Linux-only, so each fix has a POSIX-simulated regression test that forces the Windows condition — monkeypatch `Path.unlink`/`os.unlink` to raise `PermissionError`, and `Path.write_text` to do `\n`→`\r\n`. All three were verified RED (reproducing the exact bug) before the fix.

**Out of scope (follow-up):** long file names still silently skip on Windows — `_sanitize_stem` doesn't truncate, and `ee22914`'s deeper `staging/prepare/<idx>-<stem>-<uuid>/` path pushes borderline names over `MAX_PATH` (260). The root cause predates this PR, so it's left for a separate change with a proper truncation policy.

gwokhou:
## Merge conflict resolution — synced with `main`

Resolved conflicts against current `main` (`ff54396`) so this PR merges cleanly. Two files conflicted; both keep the PR's lock-free prepare/commit split intact while adopting `main`'s newer changes.

### `openkb/converter.py`
`main` moved on two axes that overlapped the PR's refactor:
- **Config loading** — `main` switched `load_config(openkb_dir/"config.yaml")` → `resolve_effective_config(kb_dir)[0]` (global-default + per-KB layering, from the effective-config work).
- **markitdown import** — `main` made `markitdown` lazy (inside the `else` branch) to keep `import openkb` cheap (#186).

**Resolution**: kept the PR's lock-free `_convert_document_impl` structure (no `kb_ingest_lock` inside, `resolve_final_name` param, hoisted config/registry) and adopted both of `main`'s changes on top — `resolve_effective_config(kb_dir)[0]` for config (it's read-only, so the lock-free prepare contract is preserved; the prepare-time `is_known` read is benign since commit re-decides skip under the lock) and the lazy `from markitdown import MarkItDown` in the `else` branch. `kb_ingest_lock` is still imported and used by the serial `convert_document` wrapper.

### `openkb/cli.py`
Both this PR and `main` added a keyword arg to `_add_single_file_locked` after `stage`:
- PR: `prepared=None` (the prepare/commit fast-path)
- `main`: `bundle=None` (per-KB REST API credential bundle)

**Resolution**: kept **both** → `stage=True, prepared=None, bundle=None`. The two flows are orthogonal: `prepared` drives the prepare/commit fast-path (re-validate skip under the lock, retarget staging to the final name, re-convert on TOCTOU); `bundle` only gates `_setup_llm_key` (`if bundle is None:`). `add_single_file` threads `bundle=` through; `commit_prepared_document` passes `prepared=` and deliberately omits `bundle` so the CLI credential setup still runs.

### Verification
- `ruff check .` ✅
- `mypy openkb` ✅ (57 files)
- `pytest` — full suite **1269 passed**, incl. `test_add_prepare.py` (private staging, commit/reaper, stale-prepare TOCTOU) and `test_add_command.py`.

No behavior change vs. the original PR — this commit is purely conflict resolution against `main`.

gwokhou:
This PR is ready for review, thank you! @KylinMountain

PR #177, opened 2026-07-09
fix(agent): make parallel_tool_calls configurable to fix Amazon Bedrock query/chat (#175)

### LINKED ISSUE #175: Issue when using Amazon Bedrock
Hi team,

I'm not sure whether my configuration for Bedrock is correct, which I believe is not fully covered in the documentation/examples.

In `<kb>/.openkb/config.yaml`, I currently have:

```
language: en
model: bedrock/eu.anthropic.claude-sonnet-4-6
aws_access_key_id: os.environ/AWS_ACCESS_KEY_ID
aws_secret_access_key: os.environ/AWS_SECRET_ACCESS_KEY
aws_region_name: os.environ/AWS_REGION_NAME
pageindex_threshold: 20
```

In `<kb>/.env`, I have:

```
AWS_ACCESS_KEY_ID=[REDACTED]
AWS_SECRET_ACCESS_KEY=[REDACTED]
AWS_REGION_NAME=eu-central-1
LLM_API_KEY=bedrock-api-key-[REDACTED]
```

When executing `query` or `chat`, I get the following error:

```
[ERROR] litellm.BadRequestError: BedrockException - {"message":"The model returned the following errors: tool_choice.type: Field required"}
```

I found an open issue at https://github.com/BerriAI/litellm/issues/27184.

1. Is my configuration correct? For instance, I'm not sure whether `LLM_API_KEY` is necessary here, or whether it's enough with the access key/secret access key pair.
2. Is the error I'm getting indeed a LiteLLM issue, or something that can be fixed in OpenKB?

KylinMountain:
Fixes #175 — Amazon Bedrock Claude models fail on every `query` / `chat` with:

```
litellm.BadRequestError: BedrockException - {"message":"The model returned the following errors: tool_choice.type: Field required"}
```

## Root cause (traced through litellm source)

`build_query_agent` hardcoded `ModelSettings(parallel_tool_calls=False)`. LiteLLM's Bedrock Converse transform turns **any** `parallel_tool_calls` value into a `tool_choice` object carrying only `{"disable_parallel_tool_use": ...}` — **with no `type` field** (`converse_transformation.py`, the `_parallel_tool_use_config` merge, gated on the model supporting parallel-tool-use config, which `claude-sonnet-4-6` does). Bedrock's Anthropic backend requires `tool_choice.type`, so it rejects the request.

- It only hits `query`/`chat` because those are the tool-calling agents; `add`/compile use plain completions (no `tool_choice`).
- Confirmed **still unfixed in the latest litellm 1.91.1** (downloaded the wheel and checked — the malformed-dict construction is unchanged; `eu.anthropic.claude-sonnet-4-6` is flagged `supports_parallel_tool_use_config: true`, so the buggy path fires). BerriAI/litellm#27184 is still open. `drop_params: true` doesn't help (Bedrock lists `parallel_tool_calls` as *supported*). So bumping litellm is not a fix.

## Fix — expose it as a config knob (default preserves current behavior)

`parallel_tool_calls` becomes a tri-state config key:

| value | effect |
|---|---|
| `false` **(default)** | force sequential tool calls — unchanged historical behavior for every provider |
| `true` | allow parallel tool calls |
| `null` | **don't send the setting** (use the provider default) |

`null` is the only value that works on Bedrock (sending any value trips the LiteLLM bug), so it's the Bedrock escape hatch. There is **no provider special-casing in code** — behavior is driven purely by the config value.

Wiring: `config.resolve_parallel_tool_calls()` (validates true/false/null, warns on junk) → `set/get_parallel_tool_calls()` process-wide stash set by `cli._setup_llm_key` and read in `build_query_agent` — same mechanism already used for `extra_headers`/`timeout`.

## Config for Bedrock (this also adds the Bedrock docs the repo lacked)

```yaml
# <kb>/.openkb/config.yaml
model: bedrock/eu.anthropic.claude-sonnet-4-6
parallel_tool_calls: null    # required for Bedrock — see #175
```
```bash
# <kb>/.env  — AWS creds do the auth; LLM_API_KEY is not needed
AWS_ACCESS_KEY_ID=...
AWS_SECRET_ACCESS_KEY=...
AWS_REGION_NAME=eu-central-1
```

## Testing

- New: resolver tests (absent→false default, true/false, explicit null→None omit, junk→default+warn), stash roundtrip, and query-agent wiring tests (stash null→omitted, false→sequential, true→parallel).
- Full suite green (951 passed); ruff check/format clean.

@cgrevisse — this should fix it. To test this branch on your Bedrock setup, add **`parallel_tool_calls: null`** to your `.openkb/config.yaml` (that tells OpenKB not to send the param LiteLLM mangles). Would appreciate a confirm! 🙏

### DIFF
--- a/config.yaml.example
+++ b/config.yaml.example
@@ -2,6 +2,12 @@ model: gpt-5.4                   # LLM model (any LiteLLM-supported provider)
 language: en                     # Wiki output language
 pageindex_threshold: 20          # PDF pages threshold for PageIndex
 
+# Optional: whether query/chat agents may call tools in parallel. Omit (the
+# default) to let the provider decide. Leave it unset for Amazon Bedrock Claude
+# models — sending it makes LiteLLM emit a malformed tool_choice and every
+# query/chat fails. Set false to force sequential tool calls on other providers.
+# parallel_tool_calls: false
+
 # Optional: override the entity-type vocabulary used for entity pages.
 # Omit this key to use the default 7 types
 # (person, organization, place, product, work, event, other).
--- a/config.yaml.example
+++ b/config.yaml.example
@@ -2,11 +2,13 @@ model: gpt-5.4                   # LLM model (any LiteLLM-supported provider)
 language: en                     # Wiki output language
 pageindex_threshold: 20          # PDF pages threshold for PageIndex
 
-# Optional: whether query/chat agents may call tools in parallel. Omit (the
-# default) to let the provider decide. Leave it unset for Amazon Bedrock Claude
-# models — sending it makes LiteLLM emit a malformed tool_choice and every
-# query/chat fails. Set false to force sequential tool calls on other providers.
-# parallel_tool_calls: false
+# Optional: whether query/chat agents may call tools in parallel.
+#   false (default) force sequential tool calls
+#   true            allow parallel tool calls
+#   null            don't send the setting (use the provider default) — set this
+#                   for Amazon Bedrock Claude, which rejects the request when
+#                   parallel_tool_calls is sent at all (any value). See #175.
+# parallel_tool_calls: null
 
 # Optional: override the entity-type vocabulary used for entity pages.
 # Omit this key to use the default 7 types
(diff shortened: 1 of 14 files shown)


cgrevisse:
Working like a charm, thanks for the quick fix!

PR #228, opened 2026-08-26
fix(web): show recent session times in local timezone

lwylab:
## Summary
- Home recent-session cards stripped the `Z` from UTC ISO timestamps and printed the digits as wall-clock time, so users in China (UTC+8) saw times ~8 hours behind.
- Parse as an absolute Instant and format with `toLocaleString` using the browser local timezone; UI language only selects locale style (`zh-CN` / en), not a hard-coded geographic zone.

## Note
Replaces closed PR #227 — GitHub refused to reopen it because `lwylab/OpenKB` is a private non-fork and head refs are not readable. This PR is opened from a fresh fork `lwylab/OpenKB-1`.

## Test plan
- [ ] Open Workbench Home with at least one recent chat session
- [ ] Confirm the card timestamp matches local clock (not UTC / ~8h behind in China)
- [ ] Switch UI language between zh and en; timezone should stay local, only date/number style may change
- [ ] Session with missing/invalid `updated_at` still renders without crashing

### DIFF
--- a/frontend/src/pages/Home.tsx
+++ b/frontend/src/pages/Home.tsx
@@ -17,13 +17,33 @@ interface RecentSession extends ChatSessionItem {
   kbIndex: number
 }
 
-function formatWhen(iso: string): string {
+/**
+ * Format a session timestamp for the recent-sessions cards.
+ *
+ * Backend stores UTC (`…Z`, see `openkb.agent.chat_session._utcnow_iso`).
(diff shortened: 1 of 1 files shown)


lwylab:
Friendly bump — this timezone display fix has been waiting ~3 weeks without review.

Would a maintainer have a chance to review when convenient?
cc @rejojer @KylinMountain

PR #250, opened 2026-09-08
fix(indexer): close PageIndex's local SQLite connection and forward LLM credentials

### LINKED ISSUE #219: [Bug] PageIndex long document indexing fails with custom OPENAI_API_BASE / 401 Unauthorized
- Description: When configuring a custom OpenAI-compatible proxy via OPENAI_API_BASE and OPENAI_API_KEY in .env, short documents process fine, but long documents (>20 pages) fail during index_long_document with litellm.AuthenticationError: 401 Unauthorized or API key not configured.

- Root Cause: openkb.indexer.index_long_document instantiates PageIndexClient without passing through the resolved base_url and api_key from the credential bundle, and pageindex.index.utils.llm_completion does not inject api_base/api_key into litellm.completion().

- Expected Behavior: PageIndexClient and its internal LiteLLM calls should inherit the active KB's resolved OPENAI_API_BASE and OPENAI_API_KEY.

### Steps to Reproduce
1. Configure `.env` with a custom OpenAI-compatible proxy:
```   
 
   OPENAI_API_BASE="http://your-proxy:port/v1"
   OPENAI_API_KEY="***"
```   
2. Set `model: openai/your-model` in `config.yaml`.
3. Ingest a long PDF (>20 pages, e.g., 30 pages) via `openkb add doc.pdf` or watch service.
4. `index_long_document` triggers `PageIndexClient`, which throws `litellm.AuthenticationError: Error code: 401 - {'error': 'API key required for remote API access'}` because credentials are not forwarded to LiteLLM inside `pageindex.index.utils`.

### LINKED ISSUE #249: fix(indexer): close PageIndex's local SQLite connection so rollback doesn't hit WinError 32
### Problem

`index_long_document()` in `openkb/indexer.py` (local PageIndex indexing for
long PDFs) never closes the local PageIndex SQLite connection it opens.
`client = PageIndexClient(...)` / `col = client.collection()` opens
`.openkb/pageindex.db` (WAL mode) and keeps it open for the lifetime of the
process — there is no `close()`/`finally` anywhere in the function.

If a later step in the same `add` (indexing itself, or the subsequent
`compile_long_doc()`) raises, the mutation rollback in `openkb/mutation.py`
tries to `unlink()`/rename `pageindex.db`/`-wal`/`-shm` as part of discarding
the failed add. On Windows this fails with `WinError 32` ("the process cannot
access the file because it is being used by another process"), because the
SQLite connection opened earlier in the *same* process is still holding the
file open:

```
openkb.mutation WARNING: Rolled back interrupted add journal ....json.
...
openkb.mutation WARNING: Mutation rollback failed: [WinError 32] Der Prozess
kann nicht auf die Datei zugreifen, da sie von einem anderen Prozess
verwendet wird: '...\.openkb\pageindex.db'
[ERROR] Rollback failed; mutation journal retained for recovery: ...
```

The rollback failure leaves a dirty mutation journal on disk and (per
`add_coordinator.DirtyRollbackError`) stops the rest of the batch instead of
continuing to the next file — a single failed long-document add can abort an
entire multi-file `openkb add` run.

### Reproduction

- `openkb add <folder-with-a-long-pdf>` on Windows.
- Indexing (or the subsequent compile step) fails for any reason — e.g. the
  credential-forwarding bug in #219, or any other transient error during
  `index_long_document()`/`compile_long_doc()`.
- The mutation rollback then hits the `WinError 32` above and the batch stops.

### Context

- Windows, `openkb` installed from `VectifyAI/OpenKB` (fork-based dev
  install), Python 3.12.
- The pinned `pageindex` version exposes `close()` only on the low-level
  `SQLiteStorage` its local backend holds internally (`client._backend
  ._storage`) — there is no public `close()`/context-manager API on
  `PageIndexClient`/`Collection` itself.
- Related: #219 (PageIndex's own LLM calls not receiving the resolved
  credential bundle) touches the same function and can be the trigger that
  causes indexing to fail and hit this rollback bug — but the two are
  independent problems (this one reproduces with any failure after PageIndex
  has opened its local DB, not just an auth failure).

### Vorschlag

Always close PageIndex's local storage connection(s) in a `finally` around
`index_long_document()`'s body, in both the success and the failure path, so
a subsequent mutation rollback can always remove `pageindex.db` on Windows —
best-effort (never raises), since cleanup must not mask the real indexing
error.

---

sebastianbraun25:
> [!NOTE]

### Problem

`index_long_document()` in `openkb/indexer.py` (local PageIndex indexing for
long PDFs) has two related robustness problems observed during a real batch
`openkb add` run:

1. The local PageIndex SQLite connection it opens (`.openkb/pageindex.db`) is
   never closed. If a later step in the same `add` fails, the mutation
   rollback in `openkb/mutation.py` tries to `unlink()`/rename
   `pageindex.db`/`-wal`/`-shm` — on Windows this hits `WinError 32` because
   the connection opened earlier in the same process is still holding the
   file open, leaving a dirty mutation journal and aborting the rest of the
   batch (#249).
2. PageIndex's own internal LLM calls (TOC/tree/summary generation) don't
   receive the KB's resolved model credentials. `index_long_document()` only
   reads `PAGEINDEX_API_KEY` (PageIndex Cloud auth) — a custom
   `LLM_API_KEY`/gateway `base_url` (the same credentials `compiler.py`'s own
   `_llm_call`/`_llm_call_async` use) never reaches PageIndex, which falls
   back to LiteLLM's default provider-key/env-var lookup instead (#219).

### Root Cause

1. `col = client.collection()` opens a WAL-mode SQLite connection via
   PageIndex's local backend, and nothing in `index_long_document()` ever
   closes it — the connection lives for the rest of the process.
2. `index_long_document()` never calls `openkb.config
   .resolve_credential_bundle()`, so it has no `api_key`/`base_url` to give
   PageIndex in the first place. Separately, the pinned `pageindex` version
   only accepts LLM credentials for its own calls via a dedicated
   `IndexConfig(llm_params={...})` field (scoped per-call through
   `pageindex.config.llm_params_scope`, context-isolated so it's safe under
   concurrent multi-KB use) — not via `PageIndexClient(api_key=...)`, which is
   PageIndex *Cloud*'s own auth and unrelated to the underlying model's LLM
   credentials in local mode.

### Solution / Changes

- `openkb/indexer.py`:
  - New `_close_pageindex_client()` helper: best-effort closes
    `client._backend._storage` (the only place the pinned `pageindex` exposes
    a `close()` — there's no public API on `PageIndexClient`/`Collection`
    itself). A no-op in cloud mode (no local backend/storage). Never raises.
  - `index_long_document()`'s whole body now runs inside a `try`/`finally`
    that calls the helper above — closed on both the success and the failure
    path.
  - `index_long_document()` now resolves `resolve_credential_bundle(kb_dir)`
    and forwards it into `_build_index_config()`.
  - `_build_index_config()` takes an optional `bundle` and forwards its
    non-empty `api_key`/`base_url` as `IndexConfig(llm_params={...})`, guarded
    by the same `IndexConfig.model_fields` check already used for
    `max_concurrency` — degrades gracefully (with a warning) against an older
    pinned PageIndex that predates the `llm_params` field.
- `tests/test_indexer.py`:
  - `TestBuildIndexConfigLlmParams`: `bundle` → `llm_params` forwarding
    (full/partial/empty/`None` bundle, and the unsupported-version fallback).
  - `TestClosePageindexClient`: closes the local backend's storage, is a
    no-op for a client without one, and swallows a `close()` exception.
  - `TestIndexLongDocument`: closes the client on both success and failure,
    and the resolved credential bundle reaches the *real* `IndexConfig`
    passed to `PageIndexClient` (not just the isolated `_build_index_config`
    unit tests).

Backward compatible: no config/CLI surface changes. A KB with no custom
`LLM_API_KEY`/`base_url` sees no `llm_params` at all (unchanged behavior).

### Issues

Resolves #249.
Resolves #219.

### DIFF
--- a/openkb/indexer.py
+++ b/openkb/indexer.py
@@ -163,14 +163,17 @@ def _build_index_config(config: dict[str, Any], bundle=None) -> IndexConfig:
     working against a pinned PageIndex that predates it (``IndexConfig``
     forbids unknown kwargs).
 
-    ``bundle``'s ``api_key``/``base_url`` (the same credentials ``compiler.py``'s
-    own LLM calls use — see :func:`openkb.config.resolve_credential_bundle`) are
-    forwarded as PageIndex's own per-call ``llm_params`` (see #219): without
-    this, PageIndex's internal indexing calls (TOC/tree/summary generation)
-    fall back to LiteLLM's default provider-key/env-var lookup, which doesn't
-    know about a KB's custom ``LLM_API_KEY``/gateway ``base_url``. Guarded by
-    the same ``model_fields`` check as ``max_concurrency`` above, so it
-    degrades gracefully against an older pinned PageIndex.
+    ``bundle``'s ``api_key``/``base_url``/``extra_headers``/``timeout`` (the
+    same credentials ``compiler.py``'s own LLM calls use — see
+    :func:`openkb.config.resolve_credential_bundle`) are forwarded as
+    PageIndex's own per-call ``llm_params`` (see #219): without this,
+    PageIndex's internal indexing calls (TOC/tree/summary generation) fall
+    back to LiteLLM's default provider-key/env-var lookup, which doesn't know
+    about a KB's custom ``LLM_API_KEY``/gateway ``base_url`` — or, for
+    gateways authenticated purely via a header (e.g. an ``Authorization:
+    Bearer`` proxy token in ``litellm.extra_headers``), has no credentials at
+    all. Guarded by the same ``model_fields`` check as ``max_concurrency``
+    above, so it degrades gracefully against an older pinned PageIndex.
     """
     kwargs: dict[str, Any] = {
         "if_add_node_text": True,
@@ -189,8 +192,13 @@ def _build_index_config(config: dict[str, Any], bundle=None) -> IndexConfig:
     if bundle is not None:
         llm_params = {
             key: value
-            for key, value in {"api_key": bundle.api_key, "base_url": bundle.base_url}.items()
-            if value
+            for key, value in {
+                "api_key": bundle.api_key,
+                "base_url": bundle.base_url,
+                "extra_headers": bundle.extra_headers,
+                "timeout": bundle.timeout,
+            }.items()
+            if value or (key == "timeout" and value is not None)
         }
         if llm_params:
             if "llm_params" in IndexConfig.model_fields:
--- a/openkb/indexer.py
+++ b/openkb/indexer.py
@@ -11,7 +11,7 @@ from typing import Any
 
 from pageindex import IndexConfig, PageIndexClient
 
-from openkb.config import resolve_concurrency, resolve_effective_config
+from openkb.config import resolve_concurrency, resolve_credential_bundle, resolve_effective_config
 from openkb.tree_renderer import render_summary_md
 
 logger = logging.getLogger(__name__)
@@ -153,7 +153,7 @@ def _write_long_doc_artifacts(
     return summary_path
 
 
-def _build_index_config(config: dict[str, Any]) -> IndexConfig:
+def _build_index_config(config: dict[str, Any], bundle=None) -> IndexConfig:
     """Build the PageIndex ``IndexConfig`` for local indexing.
 
     Forwards the KB's ``concurrency`` setting to PageIndex, which caps how many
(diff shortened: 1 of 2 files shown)


muhammadyaseen:
I ran into the credential-forwarding problem addressed by [this PR](https://github.com/VectifyAI/OpenKB/pull/250/changes/86a0539be5cdfd4deaae425435958a2acba57edf) while using OpenKB with a LiteLLM Gateway (using Azure-hosted models) protected by Cloudflare Access.

  The gateway requires these additional headers on every LLM request:

  ```yaml
  extra_headers:
    CF-Access-Client-Id: "<secret>"
    CF-Access-Client-Secret: "<secret>"
```

OpenKB already forwarded `extra_headers` for the normal compiler path, so short documents could already use the gateway. However, long PDFs routed through PageIndex did not receive the configured headers and the internal LiteLLM calls therefore could not authenticate with the gateway.

As a local workaround, I patched openkb/indexer.py to forward the resolved headers into PageIndex:

```python
from pageindex import IndexConfig, PageIndexClient, set_llm_params
from openkb.config import (
    load_config,
    resolve_concurrency,
    resolve_extra_headers,
)

# Inside index_long_document():
config = load_config(openkb_dir / "config.yaml")
set_llm_params(extra_headers=resolve_extra_headers(config))
```

The configured Cloudflare headers then appeared in PageIndex’s LiteLLM parameters, resolving the issue. I therefore wanted to support the implementation in this PR of passing `api_key, base_url, extra_headers` to PageIndex as well.

I tested against OpenKB 0.4.5 with long PDFs exceeding pageindex_threshold. No credential values are included here.

